PR #75, opened 2020-06-08
SW-427-shopware-autocomplete-when-findologic-enabled

zaifastafa:


src/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriber.php
```diff
@@ -194,10 +194,11 @@ private function handleFilters(ProductListingCriteriaEvent $event): void
             }
             $responseParser = ResponseParser::getInstance($response);
             $event->getCriteria()->addExtension('flFilters', $responseParser->getFiltersExtension());
-        } catch (ServiceNotAliveException | UnknownCategoryException $e) {
+        } catch (ServiceNotAliveException $e) {
             /** @var FindologicEnabled $flEnabled */
             $flEnabled = $event->getContext()->getExtension('flEnabled');
             $flEnabled->setDisabled();
+        } catch (UnknownCategoryException $ignored) {
```
TheKeymaster, line 201: Please add a comment, why we do not disable Findologic here. Otherwise we may remove this in the future by accident.

PR #78, opened 2020-06-16
SW-456_make_plugin_better_extandable

TheKeymaster:


src/Struct/FindologicProduct.php
```diff
@@ -91,6 +92,9 @@ class FindologicProduct extends Struct
     /** @var Property[] */
     protected $properties;
 
+    /** @var Item */
+    protected $item;
```
TobiasGraml11, line 96: It would make sense to also change all the `private` setters like `setVariantPrices` to `protected`.

PR #81, opened 2020-06-29
SW-462_allow_routerinterfaces_in_export

dblaichinger:


src/Controller/ExportController.php
```diff
@@ -35,6 +35,7 @@
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\Routing\RouterInterface;
```
TheKeymaster, line 38: I guess we can remove the old `Router` import?

dblaichinger: `Route` and `RouteScope` was unused according to IDE, removed it

src/Controller/ExportController.php
```diff
@@ -25,7 +25,6 @@
 use Shopware\Core\Framework\DataAbstractionLayer\Search\EntitySearchResult;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\IdSearchResult;
-use Shopware\Core\Framework\Routing\Annotation\RouteScope;
```
TheKeymaster, line 28: This should not be removed, as it used in a Symfony annotation. Please install the Symfony Plugin for your IDE. With it this won't count as "not imported". Also the tests break due to this :).

dblaichinger: After Symfony plugin was installed, nothing seems unused according to IDE so I just reverted the last commit.

PR #82, opened 2020-06-29
SW-451-filter-value-should-not-conflict

zaifastafa:


src/Findologic/Request/Handler/FilterHandler.php
```diff
@@ -106,7 +104,12 @@ protected function handleFilter(
         }
 
         if (in_array($filterName, $availableFilterNames, true)) {
-            $searchNavigationRequest->addAttribute($filterName, $filterValue);
+            // This resolves the SW-451 issue about filter value conflict in storefront
```
TheKeymaster, line 107: Upon click-testing I found out that the category filter does no longer work.
Clicking a category filter

1. Does not disable any other values.
1. Does not filter anything (and the URL is not changed at all).

There seems to be a console error, which is thrown on the initial load:

image

src/Findologic/Request/Handler/FilterHandler.php
```diff
@@ -174,4 +177,19 @@ private function isMaxRangeSlider(string $name): bool
...
+    {
+        return mb_strpos($filterValue, sprintf('%s-', $filterName)) === 0;
+    }
+
+    private function handlePropertyFilter(
+        string $filterName,
+        string $filterValue,
+        SearchNavigationRequest $searchNavigationRequest
+    ): void {
+        $parsedFilterValue = explode('-', $filterValue);
```
TheKeymaster, line 191: * What happens if my filter value is `some-value`? This would cause the query parameter to be `filtername-some-value`.
* What happens if my filter name is `some-filter`? This would cause the query parameter to be `some-filter-value`.

Add tests that cover that this does not cause any issues.

src/Findologic/Response/Xml21/Filter/Values/FilterValue.php
```diff
@@ -18,11 +21,14 @@ class FilterValue extends Struct
     /** @var TranslatedName */
     private $translated;
 
-    public function __construct(string $id, string $name)
+    public function __construct(string $id, string $name, ?string $filterName = null)
```
TheKeymaster, line 24: When would a filter name be null? This looks like a case that should never happen, as we always should use the internal filter name.

src/Resources/app/storefront/dist/storefront/js/fin-search.js
```diff
@@ -1 +1 @@
...
...Selector(document,this.options.parentFilterPanelSelector);this.listing=window.PluginManager.getPluginInstanceFromElement(e,"Listing"),this.listing.registerFilter(this),this._preventDropdownClose()}},{key:"_preventDropdownClose",value:function(){var e=r.a.querySelector(this.el,this.options.dropdownSelector,!1);e&&e.addEventListener("click",(function(e){e.stopPropagation()}))}},{key:"_validateMethods",value:function(){if("function"!=typeof this.getValues)throw new Error("[".concat(this._pluginName,'] Needs the method "getValues"\''));if("function"!=typeof this.getLabels)throw new Error("[".concat(this._pluginName,'] Needs the method "getLabels"\''));if("function"!=typeof this.reset)throw new Error("[".concat(this._pluginName,'] Needs the method "reset"\''));if("function"!=typeof this.resetAll)throw new Error("[".concat(this._pluginName,'] Needs the method "resetAll"\''))}}])&&c(n.prototype,o),i&&c(n,i),t}(o.a);h={parentFilterPanelSelector:".cms-element-product-listing-wrapper",dropdownSelector:".filter-panel-item-dropdown"},(p="options")in(f=y)?Object.defineProperty(f,p,{value:h,enumerable:!0,configurable:!0,writable:!0}):f[p]=h}},[["7X2L","runtime","vendor-node","vendor-shared"]]]);
```
TheKeymaster, line 1: How can this be changed if there were no JavaScript changes at all?

src/Struct/Config.php
```diff
@@ -159,7 +159,6 @@ private function initializeReadonlyConfig(?string $salesChannelId): void
     }
 
     /**
-     * @param mixed $default
```
TheKeymaster, line 162: That docblock information shouldn't get lost. Otherwise the IDE assumes that you can only add `null` here, which is not the case.

src/Findologic/Response/Xml21/Filter/Values/FilterValue.php
zaifastafa: This is null because we are not setting the ID for all filter values. Just the ones where we need it to be. For e.g the category filter does not work with IDs, as we are already having a parent category name in the value with an `_` to make it unique. The uuid is generated only for the values in which we need a unique ID for it.

src/Resources/app/storefront/dist/storefront/js/fin-search.js
zaifastafa: This was changed because I was testing different things in terms of extending a JS plugin. So it might have compiled it.

src/Struct/Config.php
zaifastafa: I think this was done by the linter. But OK I will revert this.

src/Findologic/Response/Xml21/Filter/Values/FilterValue.php
TheKeymaster: I see. Please add a docblock why `$filterName` is necessary. This way future developers understand why you sometimes need the name of the filter and sometimes not.

src/Resources/app/storefront/dist/storefront/js/fin-search.js
TheKeymaster: Please revert this then.

zaifastafa:
@TheKeymaster do not review this yet. I think there are some changes needed so I will push the code again and let you know.

src/Findologic/Request/Handler/FilterHandler.php
```diff
@@ -13,7 +14,6 @@
 class FilterHandler
 {
     protected const FILTER_DELIMITER = '|';
```
TheKeymaster, line 16: I am pretty sure if you select multiple filters, they are separated with this character already. Please ensure that when configuring a filter as "multi select" in the filter configuration, that it still works properly.

Also a character like < or > may be chosen, since you can not configure these as filter names in the Shopware backend.

src/Findologic/Response/Xml21/Filter/Values/FilterValue.php
```diff
@@ -18,11 +23,17 @@ class FilterValue extends Struct
     /** @var TranslatedName */
     private $translated;
 
-    public function __construct(string $id, string $name)
+    public function __construct(string $id, string $name, string $filterName = null)
     {
         $this->id = $id;
         $this->name = $name;
         $this->translated = new TranslatedName($name);
+        // This can be null because we do not want to set this for all filter values.
```
TheKeymaster, line 31: Put this information to the docblock of the method argument.

tests/Findologic/Request/FilterHandlerTest.php
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * @dataProvider filterRequestProvider
+     */
+    public function testFilters(string $filterName, string $filterValue, string $expectedValue)
+    {
+        $searchNavigationRequest = new SearchRequest();
+        $expectedColorFilter = new ColorPickerFilter($filterName, $filterName);
+        $expectedColorFilter->addValue(
+            (new ColorFilterValue($filterValue, $filterValue, $filterName))
```
TheKeymaster, line 57: These brackets look weirdly indented and at wrong places. Somehow at `setMedia` there are two closing brackets.

tests/Findologic/Response/Xml21ResponseParserTest.php
```diff
@@ -169,52 +169,52 @@ public function filterResponseProvider(): array
 
         $expectedVendorFilter = new VendorImageFilter('vendor', 'Hersteller');
         $expectedVendorFilter->addValue(
-            (new ImageFilterValue('Anderson, Gusikowski and Barton', 'Anderson, Gusikowski and Barton'))
+            (new ImageFilterValue('Anderson, Gusikowski and Barton', 'Anderson, Gusikowski and Barton', 'vendor'))
```
TheKeymaster, line 172: Set the filter names to a variable. This way if we change it in the future you do not have to change it manually multiple times in the same methods.

src/Findologic/Request/Handler/FilterHandler.php
```diff
@@ -174,4 +178,19 @@ private function isMaxRangeSlider(string $name): bool
...
+    {
+        return mb_strpos($filterValue, sprintf('%s%s', $filterName, FilterValue::DELIMITER)) === 0;
+    }
+
+    private function handlePropertyFilter(
+        string $filterName,
+        string $filterValue,
+        SearchNavigationRequest $searchNavigationRequest
+    ): void {
+        $parsedFilterValue = explode(sprintf('%s%s', $filterName, FilterValue::DELIMITER), $filterValue);
```
TheKeymaster, line 192: * What happens if a filter name is `Best|Filter|Name`?
* What happens if a filter value is `Best|Filter|Value`?

Does Shopware prevent this?

src/Findologic/Request/Handler/FilterHandler.php
zaifastafa: I have used the `~` for the unique filter value delimiter. It is in the `FilterValue` class. This one is for the multi-select yes which we are handling already.

zaifastafa: And we are not modifying the filter name in this story. Just adding the filter name in front of the existing value to make it unique. So the rest of the functionality should still work.

src/Findologic/Request/Handler/FilterHandler.php
zaifastafa: yes that would cause some problems I guess. Although am not sure if shopware allows that or not yet.

zaifastafa: But I guess that is not related to story, its a general logical issue I think. Because shopware also parses multiple values with `|` so it could be considered a bug in shopware if it allows a filter name to be that way. We might have to parse the values otherwise. We can discuss this in tomorrow's call.

src/Utils/Utils.php
```diff
@@ -147,4 +146,23 @@ public static function isStagingSession(Request $request): bool
 
         return false;
     }
+
+    public static function removeSpecialCharacters($name, $replace = '-')
```
TheKeymaster, line 150: That method is not used anywhere.

src/Utils/Utils.php
```diff
@@ -147,4 +146,23 @@ public static function isStagingSession(Request $request): bool
...
+    {
+        $name = iconv('utf-8', 'ascii//translit', $name);
+        $name = preg_replace('#[^A-z0-9\-_]#', $replace, $name);
+        $name = preg_replace('#-{2,}#', $replace, $name);
+        $name = trim($name, $replace);
+
+        return mb_substr($name, 0, 180);
+    }
+
+    public static function clean($string)
```
TheKeymaster, line 160: That method is not used anywhere.

PR #83, opened 2020-07-01
SW-463_fix_filter_name

TheKeymaster:


src/Utils/Utils.php
```diff
@@ -43,6 +42,16 @@ public static function removeControlCharacters(?string $string): string
         return $result ?? $string;
     }
 
+    /**
+     * Filter names are not allowed to have special characters inside of them. If they do, you may not be able
+     * to properly select a filter in certain scenarios. This is only relevant for internal naming, which is
+     * used for the query parameters. The naming in the customer-login can differ though.
```
dblaichinger, line 48: > A filter name should not contain special characters so it can be selected properly.

I would go with this sentence only, the other information doesn't help.

src/Utils/Utils.php
```diff
@@ -43,6 +42,16 @@ public static function removeControlCharacters(?string $string): string
...
+     */
+    public static function cleanFilterName(string $filterName): string
+    {
+        return preg_replace('/[^äöü_A-Za-z0-9:-]/u', '', $filterName);
```
dblaichinger, line 52: We are allowing umlauts?

dblaichinger: Maybe reorder to `A-Za-z0-9:-_`?

tests/Struct/FindologicProductTest.php
```diff
@@ -379,6 +379,72 @@ public function testProductRatings(array $ratings, float $expectedRating): void
...
+            ],
+            'filter with brackets' => [
+                'attributeName' => 'Farbwiedergabe (Ra/CRI)',
+                'expectedName' => 'FarbwiedergabeRaCRI'
+            ],
+            'filter with special UTF-8 characters' => [
+                'attributeName' => 'Ausschnitt D ø (mm)',
+                'expectedName' => 'AusschnittDmm'
+            ],
+            'filter dots and dashes' => [
```
dblaichinger, line 397: If we need to allow umlauts please also test it

tests/Struct/FindologicProductTest.php
```diff
@@ -379,6 +379,72 @@ public function testProductRatings(array $ratings, float $expectedRating): void
...
+
+    /**
+     * @dataProvider attributeProvider
+     */
+    public function testAttributesAreProperlyEscaped(string $attributeName, string $expectedName): void
+    {
+        $productEntity = $this->createTestProduct([
+            'properties' => [
+                [
+                    'id' => Uuid::randomHex(),
```
dblaichinger, line 412: Is that really a hex?

dblaichinger: I guess `randomInt` is not available :)

tests/Struct/FindologicProductTest.php
```diff
@@ -379,6 +379,72 @@ public function testProductRatings(array $ratings, float $expectedRating): void
...
+                    'group' => [
+                        'id' => Uuid::randomHex(),
+                        'name' => $attributeName
+                    ],
+                ]
+            ]
+        ]);
+
+        $criteria = new Criteria([$productEntity->getId()]);
+        $criteria = Utils::addProductAssociations($criteria);
```
dblaichinger, line 423: Just curious, what are all those associations which get added in this method?

tests/Struct/FindologicProductTest.php
```diff
@@ -379,6 +379,72 @@ public function testProductRatings(array $ratings, float $expectedRating): void
...
+            $productEntity,
+            $this->router,
+            $this->getContainer(),
+            $this->defaultContext,
+            $this->shopkey,
+            $customerGroupEntities,
+            new XMLItem('123')
+        );
+
+        $this->assertCount(5, $findologicProduct->getAttributes());
```
dblaichinger, line 444: Why 5?

src/Struct/FindologicProduct.php
```diff
@@ -415,7 +415,7 @@ protected function getAttributeProperties(
             $group = $propertyGroupOptionEntity->getGroup();
             if ($group && $propertyGroupOptionEntity->getTranslation('name') && $group->getTranslation('name')) {
                 $properyGroupAttrib = new Attribute(
-                    Utils::removeControlCharacters($group->getTranslation('name')),
+                    Utils::cleanFilterName($group->getTranslation('name')),
```
dblaichinger, line 418: I would just go for `removeSpecialChars`, it takes a string a`preg_replace` and stuff.

tests/Struct/FindologicProductTest.php
TheKeymaster: Shopware 6 uses hex values for ids.

tests/Struct/FindologicProductTest.php
TheKeymaster: Associations are basically database joins. For products we have several like, images, properties, etc.

tests/Struct/FindologicProductTest.php
TheKeymaster: Because there are several other attributes like `cat_url` or `vendor`. All in all, this are 5 :).

src/Utils/Utils.php
TheKeymaster: HTML Element ids allow umlauts. Reordering makes sense tough.

src/Utils/Utils.php
TheKeymaster: That comment wouldn't make any sense if I would rename the method.

dblaichinger: Then please improve your current comment:

> certain scenarios

Which scenarios? This is not helpful.

> This is only relevant for internal naming, which is used for the query parameters. The naming in the customer-login can differ though

Does anybody care if naming in clogin is different?

tests/Struct/FindologicProductTest.php
dblaichinger: Where do I see all the other attributes? In the test one property gets set.

src/Utils/Utils.php
TheKeymaster: I removed the comment entirely in the last review, since it didn't make sense when the method name is `removeSpecialChars`. I am fine with that.

tests/Struct/FindologicProductTest.php
TheKeymaster: They are automatically created since the product needs to have some data to even be created in the first place. The method `ProductHelper::createTestProduct` takes care of that.

https://github.com/findologic/plugin-shopware-6/blob/96099d5de1152caba36f7bd7f1e979c5e50dff8e/tests/Struct/FindologicProductTest.php#L413

dblaichinger: Ok I understand but the numer of attributes is irrelevant for the test. Can you please change the assertion so we check if `$attributeName` is in the properties?

PR #90, opened 2020-07-28
SW-469_fix_export_of_canonical_url

dblaichinger:


src/Controller/ExportController.php
```diff
@@ -296,7 +303,7 @@ private function buildXmlProducts(
                     $productEntity,
                     $this->router,
                     $this->container,
-                    $salesChannelContext->getContext(),
+                    $this->salesChannelContext->getContext(),
```
TobiasGraml11, line 306: You can just use `$this->salesChannelContext` as a parameter, because the `$context` is never used in `FindologicProduct`.
Doing that, you can remove the `fin_search.sales_channel_context` service afterwards.

src/Resources/config/services.xml
```diff
@@ -111,5 +111,7 @@
             <argument id="service_container" type="service" />
         </service>
 
+        <service id="fin_search.sales_channel_context" synthetic="true" public="true"/>
```
TobiasGraml11, line 114: Can be removed after my previous comment is resolved

src/Struct/FindologicProduct.php
```diff
@@ -26,12 +26,15 @@
...
 use Shopware\Core\Content\Seo\SeoUrl\SeoUrlCollection;
-use Shopware\Core\Content\Seo\SeoUrl\SeoUrlEntity;
 use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
 use Shopware\Core\Framework\DataAbstractionLayer\Pricing\Price as ProductPrice;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
 use Shopware\Core\Framework\Struct\Struct;
+use Shopware\Core\System\SalesChannel\Aggregate\SalesChannelDomain\SalesChannelDomainCollection;
+use Shopware\Core\System\SalesChannel\Aggregate\SalesChannelDomain\SalesChannelDomainEntity;
```
TobiasGraml11, line 36: Both of them are not used

src/Struct/FindologicProduct.php
```diff
@@ -50,6 +53,9 @@ class FindologicProduct extends Struct
     /** @var Context */
     protected $context;
 
+    /** @var SalesChannelContext */
+    protected $salesChannelContext;
+
```
TobiasGraml11, line 58: `$context` can be replaced with `$salesChannelContext`, instead of added additionally

src/Controller/ExportController.php
TheKeymaster: No because that would break backwards compatibility.

src/Resources/config/services.xml
TheKeymaster: See [my comment](https://github.com/findologic/plugin-shopware-6/pull/90#discussion_r466236349).

src/Struct/FindologicProduct.php
TheKeymaster: That would also break backwards compatibility. Such a change may be done when we remove this parameter. For now I will add a deprecation notice for this parameter.

PR #92, opened 2020-08-04
SW-473_support_6_3_0_0

TheKeymaster:
* Fix administration configuration page
 * Fix query info message
 * Fix error when opening search result page
 * Fix display of filters


src/Findologic/Request/Handler/SearchNavigationRequestHandler.php
```diff
@@ -19,6 +19,7 @@
 use FINDOLOGIC\FinSearch\Findologic\Response\ResponseParser;
 use FINDOLOGIC\FinSearch\Struct\Config;
 use FINDOLOGIC\FinSearch\Struct\QueryInfoMessage\QueryInfoMessage;
+use FINDOLOGIC\FinSearch\Struct\SystemAware;
```
TheKeymaster, line 22: Remove this import.

PR #96, opened 2020-08-28
SW-484_do_not_export_disabled_properties

TheKeymaster:


TobiasGraml11 (approved):
LGTM!

TobiasGraml11 (changes requested):
Please export non-filterable properties as `Property` in the export

tests/Struct/FindologicProductTest.php
```diff
@@ -494,6 +494,71 @@ function (Attribute $attribute) use ($expectedName) {
...
+        $foundProperties = array_filter(
+            $findologicProduct->getProperties(),
+            function (Property $property) {
+                return $property->getKey() === 'blub';
+            }
+        );
+
+        /** @var Property $property */
+        $property = reset($foundProperties);
+        $this->assertEquals('some value', $property->getAllValues()['']); // '' = Empty usergroup.
```
dblaichinger, line 559: Also extract 'some value' as `$expectedPropertyName`

tests/Struct/FindologicProductTest.php
```diff
@@ -494,6 +494,71 @@ function (Attribute $attribute) use ($expectedName) {
...
+        $productEntity = $this->createTestProduct([
+            'properties' => [
+                [
+                    'id' => Uuid::randomHex(),
+                    'name' => 'some value',
+                    'group' => [
+                        'id' => Uuid::randomHex(),
+                        'name' => 'blub',
+                        'filterable' => false
+                    ],
```
dblaichinger, line 512: It would be good to test `'filterable' => true` as well

dblaichinger (changes requested):

tests/Struct/FindologicProductTest.php
TheKeymaster: That is the default. We already have many tests that test this behavior.

src/Struct/FindologicProduct.php
```diff
@@ -425,33 +427,97 @@ protected function getAttributeProperties(ProductEntity $productEntity): array
...
+            }
+        }
+
+        return $properties;
+    }
+
+    /**
+     * @return Attribute[]
+     */
+    protected function getAttributePropertyAsAttribute(
```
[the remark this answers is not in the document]
TheKeymaster: I thought a bit, but either we have multiple if statements, which makes this entire block really hardly readable, or we have two separate methods. Maybe you have another idea. I have already invested some time thinking about a better solution though.

src/Struct/FindologicProduct.php
```diff
@@ -425,33 +427,97 @@ protected function getAttributeProperties(ProductEntity $productEntity): array
...
+
+            $groupName = $group->getTranslation('name');
+            $optionName = $settingOption->getTranslation('name');
+            if ($groupName && $optionName) {
+                $configAttrib = new Attribute(
+                    Utils::removeSpecialChars($groupName),
+                    [Utils::removeControlCharacters($optionName)]
+                );
+
+                $attributes[] = $configAttrib;
```
dblaichinger, line 520: Is the `return` missing?

src/Struct/FindologicProduct.php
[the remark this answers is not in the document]
dblaichinger: Isn't the only difference the instances, either `new Attribute` or `new Property`? Then I suggest to extract those differences in separate methods and call the methods depending on the case

dblaichinger (changes requested):

src/Struct/FindologicProduct.php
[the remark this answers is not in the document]
dblaichinger: As discussed, we leave it for now.

dblaichinger (approved):

PR #98, opened 2020-09-01
SW-485_ignore_same_range_rating

TheKeymaster:


src/Findologic/Response/Xml21/Filter/Filter.php
```diff
@@ -211,12 +211,16 @@ private static function parseSubFilters(CategoryFilterValue $filterValue, array
         }
     }
 
-    private static function handleRatingFilter(ApiRangeSliderFilter $filter): RatingFilter
+    private static function handleRatingFilter(ApiRangeSliderFilter $filter): ?RatingFilter
     {
+        $totalRange = $filter->getAttributes()->getTotalRange();
+        if ($totalRange->getMin() === $totalRange->getMax()) {
```
dblaichinger, line 217: Why is `handleRatingFilter` called when min and max are the same? Shouldn't the method call be skipped since there is nothing to handle.

dblaichinger (changes requested):

src/Findologic/Response/Xml21/Filter/Filter.php
TheKeymaster: This method is called from `\FINDOLOGIC\FinSearch\Findologic\Response\Xml21\Filter\Filter::getInstance`. 

image

Each handler should be responsible for his own filter type.

dblaichinger: I see then I'm fine with it

dblaichinger (approved):

PR #102, opened 2020-09-23
Sw 472 fix remaining points

zaifastafa:


src/Resources/app/administration/src/module/findologic-module/components/findologic-config/index.js
```diff
@@ -1,196 +1,215 @@
...
+      return typeof value !== 'boolean'
+    },
+
+    /**
+     * @public
+     * @param prop
+     * @param order
+     * @returns {function(*, *): number}
+     */
+    compare (prop = 'name', order = 'asc') {
```
TheKeymaster, line 97: `compare` is a weird name, especially when you have an optional parameter called `order`. Actually the order parameter hides the fact that you actually just compare weither something is bigger or smaller than the other value. That method is also pretty much useless on it's own, as it is strictly tight to the `sort` method of any search result entity.

I would suggest rewriting that method to accept any repository search result, and orders them by the given order parameter. This way you do not need to handle a callback or manually need to put the callback into the sort function of the search result.

src/Resources/app/administration/src/module/findologic-module/components/findologic-config/index.js
```diff
@@ -1,196 +1,215 @@
...
+      criteria.addSorting(Criteria.sort('parentId', 'ASC'))
+
+      if (this.term) {
+        criteria.addFilter(Criteria.contains('name', this.term))
+      }
+
+      return criteria
     }
-});
+
```
TheKeymaster, line 213: Empty line.

src/Resources/app/administration/src/module/findologic-module/components/findologic-config/index.js
```diff
@@ -1,196 +1,215 @@
...
+    salesChannelRepository () {
+      return this.repositoryFactory.create('sales_channel')
+    },
+
+    categoryRepository () {
+      return this.repositoryFactory.create('category')
+
+    },
+
+    criteria () {
```
TheKeymaster, line 202: If this should be a property, please give it an appropriate name. "criteria" isn't very descriptive.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #104, opened 2020-09-29
SW-501 Fix redirect of vendor clicks in SS for shops with base path

TobiasGraml11:


src/Findologic/Request/Handler/FilterHandler.php
```diff
@@ -90,8 +90,7 @@ public function handleFindologicSearchParams(Request $request): ?string
         $params = array_merge($queryParams, $mappedParams);
 
         return sprintf(
```
TheKeymaster, line 92: With this change the `sprintf` is basically pretty useless. You could just return `'?' . http_build_query(...)`.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #108, opened 2020-10-07
Sw 483 ignore empty properties

zaifastafa:


src/Struct/FindologicProduct.php
```diff
@@ -350,70 +498,149 @@ public function getProperties(): array
...
+        }
+
+        if ($this->product->getHeight()) {
+            $value = (string)$this->product->getHeight();
+            if (!Utils::isEmpty($value)) {
+                $property = new Property('height');
+                $property->addValue($value);
+                $this->properties[] = $property;
+            }
+        }
```
TheKeymaster, line 600: These blocks appear quite often. You probably should add a method that adds the value to `$this->properties` if it's not empty.

src/Struct/FindologicProduct.php
```diff
@@ -533,8 +764,7 @@ protected function setAttributeProperties(): void
 
     protected function setAdditionalAttributes(): void
     {
-        $translationKey = $this->product->getShippingFree() ? 'finSearch.general.yes' : 'finSearch.general.no';
-        $shippingFree = $this->translator->trans($translationKey);
+        $shippingFree = $this->translateBooleanValue($this->product->getShippingFree());
```
TheKeymaster, line 767: Good helper :+1:

src/Struct/FindologicProduct.php
```diff
@@ -791,89 +909,27 @@ protected function getPricesFromProduct(ProductEntity $variant): array
...
      */
     protected function getEncodedUrl(string $url): string
     {
         $parsedUrl = parse_url($url);
-
-        $parsedUrl['path'] = implode('/', array_map(
-            '\FINDOLOGIC\FinSearch\Utils\Utils::multiByteRawUrlEncode',
-            explode('/', $parsedUrl['path'])
-        ));
+        $parsedUrl['path'] = implode(
```
TheKeymaster, line 926: Thats not well readable at all. Try to find a way to make this easier understandable.

src/Struct/FindologicProduct.php
```diff
@@ -960,16 +1004,23 @@ protected function getCustomFieldProperties(ProductEntity $product): array
...
 
         foreach ($productFields as $key => $value) {
-            if (is_string($value)) {
-                $value = Utils::cleanString($value);
+            if (!Utils::isEmpty($key) && !Utils::isEmpty($value)) {
+                if (is_string($value)) {
+                    $value = Utils::cleanString($value);
+                }
+
+                if (is_bool($value)) {
```
TheKeymaster, line 1017: :ok_hand:

tests/Controller/ExportControllerTest.php
```diff
@@ -769,4 +774,141 @@ public function testExportHeaders(array $expectedHeaders): void
...
+                'Error while exporting the product with id "%s". If you see this message in your logs, ' .
+                'please report this as a bug. Error message: Test Exception'
+            ]
+        ];
+    }
+
+    /**
+     * @dataProvider exceptionMessageProvider
+     */
+    public function testExceptionIsNotLogged(string $exceptionMessage): void
```
TheKeymaster, line 795: What is this test? I tried reading it a couple of times but it does not seem to make sense. If an exception is thrown, we want to catch it and log it. But here you expect nothing but be logged, so why would you even care about the message, if it is never called anyway? Seems odd. Also the data providers do not change how the test runs, therefore they will both return the same result.

TheKeymaster (changes requested):

src/Struct/FindologicProduct.php
```diff
@@ -923,13 +867,9 @@ protected function setProductPrices(): void
...
-        $parsedUrl['path'] = implode(
-            '/',
-            array_map(
-                '\FINDOLOGIC\FinSearch\Utils\Utils::multiByteRawUrlEncode',
-                explode('/', $parsedUrl['path'])
-            )
-        );
+        $urlPath = explode('/', $parsedUrl['path']);
+        $encodedPath = array_map('\FINDOLOGIC\FinSearch\Utils\Utils::multiByteRawUrlEncode', $urlPath);
+        $parsedUrl['path'] = implode('/', $encodedPath);
```
TheKeymaster, line 872: Nice simplification :+1:

src/Struct/FindologicProduct.php
```diff
@@ -1040,4 +980,13 @@ private function translateBooleanValue(bool $value)
 
         return $this->translator->trans($translationKey);
     }
+
+    private function addProperty(string $name, $value)
```
TheKeymaster, line 984: Please make this method `protected`, so it can be called in an extension.

TheKeymaster: Otherwise neat method :)

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #109, opened 2020-10-09
SW-510_fix_services

TheKeymaster:


src/Struct/FindologicProduct.php
```diff
@@ -686,7 +686,7 @@ protected function setCategoriesAndCatUrls(): void
                 continue;
             }
 
-            if (!strpos($categoryEntity->getPath(), $navigationCategoryId)) {
+            if (!$categoryEntity->getPath() || !strpos($categoryEntity->getPath(), $navigationCategoryId)) {
```
TheKeymaster, line 689: An exception occurred if the path has been `null`. 

Discussed this with @TobiasGraml11, since he implemented this. `null` values should be handled the same way as empty strings, therefore they're skipped.

dblaichinger (approved):

PR #111, opened 2020-10-14
SW-513_seo_url_translations

TheKeymaster:


src/Struct/FindologicProduct.php
```diff
@@ -325,6 +325,50 @@ protected function setUrl(): void
...
 
+    protected function getTranslatedSeoPath(): ?string
+    {
+        $salesChannel = $this->salesChannelContext->getSalesChannel();
+        $seoUrlCollection = $this->product->getSeoUrls()->filterBySalesChannelId($salesChannel->getId());
+
+        /** @var SeoUrlEntity|null $seoUrlEntity */
+        $seoUrlEntity = $this->getTranslatedEntity($seoUrlCollection);
+
+        return $seoUrlEntity ? $seoUrlEntity->getSeoPathInfo() : $seoUrlEntity;
```
dblaichinger, line 336: Either a string or null is returned, right? Could be more clear when you explicitly return `null` here

src/Struct/FindologicProduct.php
```diff
@@ -325,6 +325,50 @@ protected function setUrl(): void
...
+
+    protected function getTranslatedDomainBaseUrl(): ?string
+    {
+        $salesChannel = $this->salesChannelContext->getSalesChannel();
+        $domainCollection = $salesChannel->getDomains();
+
+        /** @var SalesChannelDomainEntity|null $domainEntity */
+        $domainEntity = $this->getTranslatedEntity($domainCollection);
+
+        return $domainEntity ? $domainEntity->getUrl() : $domainEntity;
```
dblaichinger, line 347: Either a string or null is returned, right? Could be more clear when you explicitly return `null` here

src/Struct/FindologicProduct.php
```diff
@@ -325,6 +325,50 @@ protected function setUrl(): void
...
+        $translatedEntities = $collection->filterByProperty(
+            'languageId',
+            $this->salesChannelContext->getSalesChannel()->getLanguageId()
+        );
+
+        if ($translatedEntities->count() === 0) {
+            return null;
+        }
+
+        return $translatedEntities->first();
```
dblaichinger, line 369: Possible to have more than one translatedEntitites?

dblaichinger (changes requested):

src/Struct/FindologicProduct.php
TheKeymaster: Theoretically yes, but practically no, as the SEO URLs are generated by Shopware via the indexer.

dblaichinger (approved):

PR #115, opened 2020-10-22
Sw 502  third party plugin search result limit

zaifastafa:


src/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriber.php
```diff
@@ -137,9 +137,16 @@ public function handleResult(ProductListingResultEvent $event): void
 
     public function handleListingRequest(ProductListingCriteriaEvent $event): void
```
TheKeymaster, line 138: These changes should also be done in the `CompatibilityLayer` classes.

src/Utils/Utils.php
```diff
@@ -223,4 +223,26 @@ public static function isEmpty($value): bool
...
+    public static function getOffset(Request $request, ?int $limit = null)
+    {
+        if (!$limit) {
+            $limit = Pagination::DEFAULT_LIMIT;
+        }
+
+        $page = self::getPage($request);
+
+        return ($page - 1) * $limit;
+    }
```
TheKeymaster, line 247: These are no utility functions, because other than in search, this is basically useless. Put these methods into a class that represents their usage.

src/Traits/SearchResultHelper.php
```diff
@@ -31,9 +31,10 @@ protected function assignPaginationToCriteria(Criteria $criteria): void
         /** @var Pagination $pagination */
         $pagination = $criteria->getExtension('flPagination');
         if ($pagination) {
-            // Pagination is handled by FINDOLOGIC.
-            $criteria->setLimit(24);
-            $criteria->setOffset(0);
+            // Pagination is handled by FINDOLOGIC. If there is an existing limit set, we respect that,
+            // otherwise use 24 as default
```
TheKeymaster, line 35: Add a `.` at the end of your sentence.

tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
```diff
@@ -893,4 +895,73 @@ private function getDefaultSessionMock(): SessionInterface
...
+            ],
+            'includes' => null
+        ];
+        if (Utils::versionLowerThan('6.3')) {
+            $expectedAssign['source'] = null;
+        }
+        $expectedAssign['title'] = null;
+
+        $criteriaMock = $this->getMockBuilder(Criteria::class)->disableOriginalConstructor()->getMock();
+        $criteriaMock->expects($this->any())->method('assign')->with($expectedAssign);
```
TheKeymaster, line 958: If you use `$this->any()` in combination with a `with()`, you are never sure that this has ever been called, and you can not be sure that the data has been properly set. I suggest using a real criteria and asserting the actual data of it after calling the subscriber search endpoint.

In fact I am sure that this test must fail, since the required change has not been added to the `CompatibilityLayer` classes, and yet the test still passes, which should not be the case.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #118, opened 2020-10-29
SW-525_hide_shopware_autocomplete_if_fl_is_active

TheKeymaster:


tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
```diff
@@ -301,7 +305,7 @@ public function testResponseHasPromotion(bool $isSearch, string $endpoint): void
 
     public function testResponseHasNoPromotion(): void
     {
-        $this->configMock->expects($this->once())->method('isActive')->willReturn(true);
+        $this->configMock->expects($this->any())->method('isActive')->willReturn(true);
```
dblaichinger, line 308: Possible to use `->exactly()` here? Or is there a reason not to use it?

tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
```diff
@@ -828,20 +841,23 @@ private function setUpSearchRequestMocks(
...
+            $context = $this->getMockBuilder(Context::class)->disableOriginalConstructor()->getMock();
+            $context->expects($this->any())->method('getExtension')->willReturnMap($defaultExtensionMap);
+
+            if ($withSmartDidYouMean) {
+                $context->expects($this->any())->method('addExtension')->withConsecutive(
+                    ['findologicService', $findologicService],
+                    ['flSmartDidYouMean', $smartDidYouMean]
+                );
+            } else {
+                $context->expects($this->any())->method('addExtension')->withConsecutive(
```
dblaichinger, line 854: Should be just `->with()` instead of `->withConsecutive()`

tests/Utils/UtilsTest.php
```diff
@@ -149,4 +156,154 @@ public function testCleanStringMethod($text, $expected, $errorMessage): void
...
+            ],
+            'Plugin is active and direct integration with staging but without findologic=on' => [
+                'isActive' => true,
+                'isActiveOnCategory' => true,
+                'isDirectIntegration' => true,
+                'isStagingShop' => true,
+                'isStagingSession' => false,
+                'isCategoryPage' => false,
+                'expectedFindologicActive' => false,
+                'expectedSmartSuggestActive' => true, // Smart Suggest handles findologic=on in case of DI.
```
dblaichinger, line 231: > // Smart Suggest handles findologic=on in case of DI.

What does that mean? 
In case Staging shop but without the flag, I would assume that this should be `false`

dblaichinger (changes requested):

tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
TheKeymaster: I only want to tell the config once that it is active, if it is active, why should it not be active in the same test? Doesn't make sense, so I explicitly wrote `any`.

tests/Utils/UtilsTest.php
TheKeymaster: If it is Direct Integration, we will let the Direct Integration decide. It is not the job of the plugin to decide weither to enable direct integration or not.

tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
TheKeymaster: True, in this case it doesn't make sense to use `withConsecutive`.

tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
dblaichinger: Alright

src/Utils/Utils.php
```diff
@@ -169,11 +165,19 @@ public static function shouldHandleRequest(
...
+            $findologicService->enableSmartSuggest();
+        }
+
+        if (!$config->isActive() || ($isCategoryPage && !$config->isActiveOnCategoryPages())) {
+            return $findologicService->disable();
+        }
+
         if ($isDirectIntegration || !$allowRequestForStaging) {
-            return $findologicService->setDisabled();
+            return $findologicService->disable();
```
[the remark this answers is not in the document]
TheKeymaster: No the Smart Suggest has an own path. It is disabled by default, we only activate it in a specific case. That would also be the case for the Findologic service, since that one is also disabled by default and only enabled if really needed. That being said, we could also just return the current state here instead of explicitly manipulating the state, as that wouldn't change the outcome.

tests/Utils/UtilsTest.php
dblaichinger: But DI should also detect that findologic should be inactive with this setting. How can this still be `true`? 
The property is by default false.

TheKeymaster: This is also the same for Shopware 5 by the way.

src/Utils/Utils.php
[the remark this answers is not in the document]
dblaichinger: Alright

tests/Utils/UtilsTest.php
dblaichinger: for clarification, which code is setting it to true?

TheKeymaster: > But DI should also detect that findologic should be inactive with this setting. How can this still be true?
The property is by default false.

We explicitly check if it is Direct Integration. In this case, we will always show the Smart Suggest, no matter what, because we shouldn't care about it.

image

dblaichinger: As discussed in call, this can be resolved.

tests/Utils/UtilsTest.php
```diff
@@ -149,4 +156,154 @@ public function testCleanStringMethod($text, $expected, $errorMessage): void
...
+                'isActive' => true,
+                'isActiveOnCategory' => false,
+                'isDirectIntegration' => false,
+                'isStagingShop' => true,
+                'isStagingSession' => true,
+                'isCategoryPage' => true,
+                'expectedFindologicActive' => false,
+                'expectedSmartSuggestActive' => true,
+            ],
+        ];
```
dblaichinger, line 253: Please add a case where isActive=false and isActiveOnCategory=true

tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
TheKeymaster: I tried using `with`, but it does not seem to work, since other tests use `withConsecutive` on the same method and using `withConsecutive` multiple times on the same method, causes them to merge together.

E.g.

```php
$someMock->expects($this->any())->method('foo')->withConsecutive('bar');
$someMock->expects($this->any())->method('foo')->withConsecutive('foobar');
```

Will actually make the first access return `foo` and the second access `foobar`. Using only `with`, won't allow this merging and it therefore does not work.

dblaichinger: That might lead to confusion but we leave it for now, same logic was already there before.

dblaichinger (approved):

PR #119, opened 2020-11-03
Implement GitHub Actions (SW-537)

TheKeymaster:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,37 @@
...
+      - master
+  pull_request:
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-versions: ['7.3']
```
dblaichinger, line 15: Shouldn't the linter run for the lowest php version we support?

dblaichinger (changes requested):

.github/workflows/lint.yml
TheKeymaster: Why should it? The linter is a packagist package, which is installed via composer. The PHP version shouldn't matter.

dblaichinger: When the version doesn't matter it's fine.

dblaichinger (approved):

TheKeymaster:
Closed in favor of #129.

PR #125, opened 2020-11-09
SW-532_fix_category_translation_check

TheKeymaster:


tests/Struct/FindologicProductTest.php
```diff
@@ -1044,6 +1044,46 @@ function (SalesChannelDomainEntity $domain) use ($salesChannel) {
...
+        $mainCatId = $this->getContainer()->get('category.repository')
+            ->searchIds(new Criteria(), Context::createDefaultContext())->firstId();
+
+        $categoryId = Uuid::randomHex();
+        $productEntity = $this->createTestProduct([
+            'categories' => [
+                [
+                    'parentId' => $mainCatId,
+                    'id' => $categoryId,
+                    'name' => ' ',
```
dblaichinger, line 1058: Is empty string also possible to configure or even null?

tests/Struct/FindologicProductTest.php
```diff
@@ -1044,6 +1044,46 @@ function (SalesChannelDomainEntity $domain) use ($salesChannel) {
...
+            $this->getContainer(),
+            $this->buildSalesChannelContext(Defaults::SALES_CHANNEL, 'http://test.at')->getContext(),
+            $this->shopkey,
+            [],
+            new XMLItem('123')
+        );
+
+        $this->assertCount(6, $findologicProduct->getAttributes());
+        $this->assertSame('cat_url', $findologicProduct->getAttributes()[0]->getKey());
+        $this->assertCount(2, $findologicProduct->getAttributes()[0]->getValues());
```
dblaichinger, line 1084: I guess those attributes are cat_urls? It would be more clear if you assert the values rather than the count.

dblaichinger (changes requested):

tests/Struct/FindologicProductTest.php
TheKeymaster: Blank or null values will be prevented.

image

dblaichinger: This means the category of the affected client is configured with one space or similar?

TheKeymaster: The category of the affected client has no translation. Setting up translations is not an easy task when dealing with them in tests, so I just assume that the configured "default language" has an empty string assigned, which results in the same.

dblaichinger: Alright

dblaichinger (approved):

PR #131, opened 2020-11-19
Sw 509 configure different shopkey language sales channel

zaifastafa:


src/Migration/Migration1605714035FinSearchConfig.php
```diff
@@ -0,0 +1,86 @@
...
+class Migration1605714035FinSearchConfig extends MigrationStep
+{
+    public function getCreationTimestamp(): int
+    {
+        return 1605714035;
+    }
+
+    public function update(Connection $connection): void
+    {
+        $sql = 'CREATE TABLE IF NOT EXISTS `finsearch_config` (
```
TheKeymaster, line 22: Use [heredoc syntax](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc) when writing a string in multiple lines.

src/Migration/Migration1605714035FinSearchConfig.php
```diff
@@ -0,0 +1,86 @@
...
+    public function updateDestructive(Connection $connection): void
+    {
+        // implement update destructive
+    }
+
+    private function insertPreviousConfigurationIfExists(Connection $connection): void
+    {
+        $connection->executeUpdate('TRUNCATE TABLE `finsearch_config`');
+
+        $sql = "SELECT LOWER(HEX(`id`)) AS `id`,
```
TheKeymaster, line 48: Use [heredoc syntax](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc) when writing a string in multiple lines.

src/Migration/Migration1605714035FinSearchConfig.php
```diff
@@ -0,0 +1,86 @@
...
+            $sql = "SELECT LOWER(HEX(`language_id`)) AS `language_id` FROM `sales_channel` WHERE `id` = UNHEX(?)";
+            $languageId = $connection->fetchColumn($sql, [$salesChannelId]);
+            if (!$languageId) {
+                continue;
+            }
+
+            $data = $config;
+            $data['language_id'] = Uuid::fromHexToBytes($languageId);
+            $data['sales_channel_id'] = Uuid::fromHexToBytes($salesChannelId);
+            $data['id'] = Uuid::randomBytes();
```
TheKeymaster, line 75: Why are we being explicit about the id? Shouldn't it be added automatically by default?

src/Resources/app/administration/src/module/findologic-module/components/findologic-config/findologic-config.html.twig
```diff
@@ -1,217 +1,106 @@
...
-        <sw-card class="sw-card--grid"
-                 :title="$tc('findologic.settingForm.config.title')">
-            <sw-container>
-                {% block findologic_credentials_card_container %}
-
-                    {% block findologic_credentials_settings %}
-                        <div class="findologic-settings-credentials-fields" v-if="actualConfigData">
-                            {% block findologic_credentials_settings_shopkey %}
-                                <sw-inherit-wrapper
+    <sw-container slot="grid" rows="auto auto">
```
TheKeymaster, line 2: 1. When you access the configuration initially, you think that it is broken. Try to add a placeholder in the dropdown like "Please choose a Sales Channel..." (DE: "Bitte wählen Sie einen Verkaufskanal").
image
1. As soon as you switch to another Sales Channel and go back to "All Sales Channel", the language is still preserved. The placeholder from above should be shown instead.
image

src/Controller/FindologicConfigController.php
```diff
@@ -0,0 +1,84 @@
...
+        return new JsonResponse(null, Response::HTTP_NO_CONTENT);
+    }
+
+    /**
+     * @Route("/api/v{version}/_action/finsearch/batch", name="api.action.finsearch.save.batch",
+     *     methods={"POST"})
+     */
+    public function batchSaveConfiguration(Request $request): JsonResponse
+    {
+        foreach ($request->request->all() as $key => $kvs) {
```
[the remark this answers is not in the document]
TheKeymaster: Also "key/value" isn't very clear here. It probably makes more sense to call this `$config` or `$options`.

src/Controller/FindologicConfigController.php
```diff
@@ -0,0 +1,84 @@
...
+        $this->saveKeyValues($salesChannelId, $languageId, $kvs);
+
+        return new JsonResponse(null, Response::HTTP_NO_CONTENT);
+    }
+
+    /**
+     * @Route("/api/v{version}/_action/finsearch/batch", name="api.action.finsearch.save.batch",
+     *     methods={"POST"})
+     */
+    public function batchSaveConfiguration(Request $request): JsonResponse
```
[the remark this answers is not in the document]
TheKeymaster: Ensure that if an error happens, the customer gets a notification. Currently the customer isn't notified and thinks everything works as expected.

src/Export/SalesChannelService.php
```diff
@@ -32,26 +33,18 @@ public function __construct(
...
-                // When the configured sales channel id is null, the shopkey is configured for all sales channels.
-                if ($systemConfigEntity->getSalesChannelId() === null) {
-                    return $currentContext;
-                }
-
-                return $this->salesChannelContextFactory->create(
-                    $currentContext->getToken(),
-                    $systemConfigEntity->getSalesChannelId()
-                );
+                return $currentContext;
```
TheKeymaster, line 47: Why was this changed? The exported language should be based on the given shopkey. Please elaborate.

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+    public function getDomain(string $salesChannelId, string $languageId): array
+    {
+        $domain = 'FinSearch';
+        $queryBuilder = $this->connection->createQueryBuilder()
+            ->select('LOWER(HEX(id))')
+            ->from('finsearch_config')
+            ->where('sales_channel_id = :salesChannelId')
+            ->andWhere('language_id = :languageId');
+
+        $domain = rtrim($domain, '.') . '.';
```
TheKeymaster, line 129: You are declaring `$domain` in line `122` as `FinSearch`. All this trimming and str_replace doesn't even do anything, so why are we doing it in the first place?

The only thing that is being done, is adding a `.` to the end. Why isn't this a constant then?

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+    {
+        return $this->load($salesChannelId, $languageId);
+    }
+
+    /**
+     * @throws InvalidDomainException
+     * @throws InvalidUuidException
+     * @throws InconsistentCriteriaIdsException
+     */
+    public function getDomain(string $salesChannelId, string $languageId): array
```
TheKeymaster, line 120: Why is a method called `getDomain` returning an array containing all configuration?

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        $criteria = new Criteria($ids);
+        /** @var SystemConfigCollection $collection */
+        $collection = $this->finSearchConfigRepository
+            ->search($criteria, Context::createDefaultContext())
+            ->getEntities();
+
+        $collection->sortByIdArray($ids);
+        $merged = [];
```
TheKeymaster, line 154: `$merged` could be anything, be more precise about your variable names.

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+        foreach ($collection as $cur) {
```
TheKeymaster, line 155: `$cur` could be anything, be more precise about your variable names.

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+            ->where('sales_channel_id = :salesChannelId')
+            ->andWhere('language_id = :languageId');
+
+        $domain = rtrim($domain, '.') . '.';
+        $escapedDomain = str_replace('%', '\\%', $domain);
+
+        $salesChannelId = Uuid::fromHexToBytes($salesChannelId);
+        $languageId = Uuid::fromHexToBytes($languageId);
+
+        $queryBuilder->andWhere('configuration_key LIKE :prefix')
```
TheKeymaster, line 135: I am a bit confused about this, since we aren't using our configuration table here, but instead use the Shopware configuration. Shouldn't this data come out of our configuration table?

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+        while ($chunk = $iterator->fetch()) {
+            $configs->merge($chunk->getEntities());
+        }
+
+        $this->configs[$key] = $this->buildFindologicConfigArray($configs);
+
+        return $this->configs[$key];
+    }
+
+    private function buildFindologicConfigArray(FinSearchConfigCollection $configs): array
```
TheKeymaster, line 278: Adding type information inside of a method doesn't make much sense.
```suggestion
    private function buildFindologicConfig(FinSearchConfigCollection $configs): array
```

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+        foreach ($configs as $config) {
+            $keys = explode('.', $config->getConfigurationKey());
+
+            $configValues = $this->getSubArray($configValues, $keys, $config->getConfigurationValue());
+        }
+
+        return $configValues;
+    }
+
+    private function getSubArray(array $configValues, array $keys, $value): array
```
TheKeymaster, line 291: Adding type information inside of a method doesn't make much sense.
```suggestion
    private function getSub(array $configValues, array $keys, $value): array
```

This method is really weird, as it requires in-depth knowledge on how to use it. I wouldn't know what "getSubArray" does, even if I would check all method arguments. Rename this method and add some kind of docblock that makes understanding this method easier.

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+            )
+        );
+
+        $criteria->addSorting(
+            new FieldSorting('salesChannelId', FieldSorting::ASCENDING),
+            new FieldSorting('id', FieldSorting::ASCENDING)
+        );
+        $criteria->setLimit(500);
+
+        $configs = new FinSearchConfigCollection();
```
TheKeymaster, line 262: Why is this even declared if it is not used? In line 273 you explicitly use this collection, to save one configuration and in the next line you return it immediately. Is there a reason why it isn't returned directly?

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+        return $merged;
+    }
+
+    /**
+     * @param array|bool|float|int|string|null $value
+     */
+    public function set(string $key, $value, string $salesChannelId, string $languageId): void
+    {
+        // reset internal cache
+        $this->configs = [];
```
TheKeymaster, line 170: What are you resetting? Why are you resetting? Why needs it to be done here? Shouldn't it be replaced afterwards? That's an odd side effect.

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+        $id = $this->getId($key, $salesChannelId, $languageId);
+        if ($value === null) {
+            if ($id) {
+                $this->finSearchConfigRepository->delete([['id' => $id]], Context::createDefaultContext());
+            }
+
+            return;
+        }
+
+        $data = [
```
TheKeymaster, line 184: This can be directly assigned to the method.

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+            'configurationValue' => $value,
+            'salesChannelId' => $salesChannelId,
+            'languageId' => $languageId,
+        ];
+        $this->finSearchConfigRepository->upsert([$data], Context::createDefaultContext());
+    }
+
+    public function delete(string $key, string $salesChannel, string $languageId): void
+    {
+        $this->set($key, null, $salesChannel, $languageId);
```
TheKeymaster, line 196: Why don't you have a separate method for delete? You could just call `$this->finSearchConfigRepository->delete([['id' => $id]], Context::createDefaultContext());` here.

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    private function validate(string $key, string $salesChannelId, string $languageId): void
+    {
+        $key = trim($key);
+        if ($key === '') {
+            throw new InvalidKeyException('key may not be empty');
```
TheKeymaster, line 203: Please capitalize sentences.

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+    private function load(string $salesChannelId, string $languageId): array
+    {
+        $key = sprintf('%s-%s', $salesChannelId, $languageId);
+
+        if (isset($this->configs[$key])) {
+            return $this->configs[$key];
+        }
+
+        $criteria = new Criteria();
+        if (method_exists($criteria, 'setTitle')) {
```
TheKeymaster, line 242: Why are we checking if this method exists, and what does it?

tests/Controller/ExportControllerTest.php
```diff
@@ -180,4 +181,36 @@ protected function parsePluginVersion(): string
...
+            [
+                'translations' => [
+                    ['name' => $expectedName, 'languageId' => $anotherLanguageId],
+                ],
+            ]
+        );
+
+        $this->enableFindologicPlugin($this->getContainer(), self::VALID_SHOPKEY, $this->salesChannelContext);
+
+        $response = $this->sendExportRequest(['productId' => $product->getId()]);
```
TheKeymaster, line 207: Send another request to ensure that if there are two shopkeys for different language, each shopkey will return the product in the proper translation.

DE: `ABCDABCDABCDABCDABCDABCDABCDABCD` => Product name: `Produkt`
EN: `1BCDABCDABCDABCDABCDABCDABCDABCD` => Product name: `Product`

tests/Controller/ExportControllerTest.php
```diff
@@ -180,4 +181,36 @@ protected function parsePluginVersion(): string
 
         return ltrim($parsed['version'], 'v');
     }
+
+    public function testCorrectTranslatedProductIsExported(): void
+    {
+        $anotherLanguageId = Uuid::randomHex();
+        $this->createLanguage($anotherLanguageId);
+        $this->salesChannelContext = $this->buildSalesChannelContext(
```
TheKeymaster, line 189: The sales channel context must be built within the controller. That's what the `SalesChannelContextService` is for, it generates the sales channel context, based on the given shopkey. This means that it must be possible to call the URL https://your-shop.co.uk/findologic?shopkey=shopkey-of-german-channel and the products will be exported in German.

tests/Controller/ExportControllerTest.php
```diff
@@ -36,7 +38,7 @@ protected function setUp(): void
     public function testExportOfSingleProduct(): void
     {
         $product = $this->createVisibleTestProduct();
-        $this->enableFindologicPlugin($this->getContainer(), self::VALID_SHOPKEY);
+        $this->enableFindologicPlugin($this->getContainer(), self::VALID_SHOPKEY, $this->salesChannelContext);
```
TheKeymaster, line 41: Dont add any sales channel context in these tests. The controller must figure out which sales channel to use, based on the given shopkey.

tests/Export/SalesChannelServiceTest.php
```diff
@@ -43,11 +45,32 @@ public function testNothingConfiguredForGivenShopkeyReturnsNull(): void
...
+        $salesChannelService = $this->getSalesChannelService();
+        $currentSalesChannel = $this->buildSalesChannelContext(
+            Defaults::SALES_CHANNEL,
+            'http://test.de',
+            null,
+            $languageId
+        );
+
+        $this->enableFindologicPlugin($this->getContainer(), $configShopkey, $currentSalesChannel);
+        $salesChannel = $salesChannelService->getSalesChannelContext($currentSalesChannel, $configShopkey);
```
TheKeymaster, line 72: Add `null` as first parameter. Otherwise this test isn't really testing anything.

tests/Traits/DataHelpers/ConfigHelper.php
```diff
@@ -56,17 +61,48 @@ private function getDefaultFindologicConfigServiceMock(
         $configServiceMock->method('get')
             ->willReturnMap(
                 [
-                    ['FinSearch.config.active', $salesChannelId, $config['active']],
-                    ['FinSearch.config.shopkey', $salesChannelId, $config['shopkey']],
-                    ['FinSearch.config.activeOnCategoryPages', $salesChannelId, $config['activeOnCategoryPages']],
-                    ['FinSearch.config.crossSellingCategories', $salesChannelId, $config['crossSellingCategories']],
-                    ['FinSearch.config.searchResultContainer', $salesChannelId, $config['searchResultContainer']],
+                    [
```
TheKeymaster, line 64: This array is just bloated with new lines. Please revert this.

TheKeymaster (changes requested):

src/Findologic/Config/FinSearchConfigCollection.php
```diff
@@ -0,0 +1,40 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\EntityCollection;
+
+/**
+ * @method void                    add(FinSearchConfigEntity $entity)
+ * @method void                    set(string $key, FinSearchConfigEntity $entity)
+ * @method FinSearchConfigEntity[]    getIterator()
+ * @method FinSearchConfigEntity[]    getElements()
+ * @method FinSearchConfigEntity|null get(string $key)
+ * @method FinSearchConfigEntity|null first()
+ * @method FinSearchConfigEntity|null last()
```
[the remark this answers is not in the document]
zaifastafa: This is how Shopware handles it.

src/Export/SalesChannelService.php
zaifastafa: The condition before the return checks for the shopkey given if the sales channel has a same shopkey or not. SInce sales channel is required for shopkey in this implementation.

src/Findologic/Config/FindologicConfigService.php
zaifastafa: This is how shopware handles it. It stores the configuration in memory, and it refreshes it when you save a configuration.

src/Findologic/Config/FindologicConfigService.php
zaifastafa: No idea, but it was introduced in later Shopware version and was failing tests for older versions.

tests/Traits/DataHelpers/ConfigHelper.php
zaifastafa: It extends the column count otherwise lint will fail. This is PSR-12. Please decide which coding standard to follow.

src/Export/SalesChannelService.php
TheKeymaster: The used sales channel should not matter. We need to get the sales channel, based on the shopkey. The only exception before this was "All Sales Channel", because if it has been configured for all Sales Channels, it did not matter. Since you now no longer can configure a shopkey for all sales channels, this condition can be removed entirely.

https://german-shop.de/findologic=shopkey=german-shopkey => Must return products in German translation.
https://german-shop.de/findologic=shopkey=english-shopkey => Must return products in English translation.
etc.

src/Findologic/Config/FindologicConfigService.php
TheKeymaster: That comment does not make much sense then. Refer to a method which does this or call that method explicitly. Also why are you clearing the configuration, instead of overriding it with the new values?

src/Findologic/Config/FindologicConfigService.php
TheKeymaster: So the error was about not setting it? Could you send me a link to the GitHub Actions build where this is failing?

tests/Traits/DataHelpers/ConfigHelper.php
TheKeymaster: This is perfect PSR-12 standard. As stated [here](https://www.php-fig.org/psr/psr-12/#47-method-and-function-calls):

> Argument lists MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one argument per line. A **single argument being split across multiple lines** (as might be the case with an anonymous function **or array**) does **not constitute splitting the argument list** itself.

src/Findologic/Config/FindologicConfigService.php
zaifastafa: We do not have an ID here, it is deleting based on key, so instead of doing the logic to get the config by key it was easier to set it to null and delete in the set method (shopware does it like that as well)

src/Findologic/Config/FindologicConfigService.php
zaifastafa: Not it was that the method was undefined or did not exist. Yeah sure let me check, it was I think maybe 6.1. Will check.

tests/Traits/DataHelpers/ConfigHelper.php
zaifastafa: So how come the Phpstorm changes this even though PSR-12 is set as the code style?

zaifastafa: It might have not complained before as there was no language ID being passed. After passing language ID, it exceeds the column length

src/Findologic/Config/FindologicConfigService.php
TheKeymaster: Okay fine then.

tests/Traits/DataHelpers/ConfigHelper.php
TheKeymaster: I think this was a bit of misunderstanding. Yes the project applies PSR-12 and yes, the maximum character limit per lines is 120. This was actually my bad as I thought you added one additional case and now you have moved all lines into this bloated structure, but it actually needs to be like that, since the language id is now also passed.
You do not have to fix anything here, sorry for the confusion ^^.

tests/Controller/ExportControllerTest.php
zaifastafa: @TheKeymaster  The sales channel was already a 3rd argument in the plugin helper trait before my changes. I am passing a sales channel because we need one to set the config as for testing it sets a given shopkey in the sales channel. Otherwise the trait has no way to check what sales channel it has.

tests/Controller/ExportControllerTest.php
zaifastafa: @TheKeymaster It is already being built in the `setUp` method, I am only creating a different one here due to a different language test case

tests/Export/SalesChannelServiceTest.php
zaifastafa: @TheKeymaster We need the context to search the repository for the shopkey. and to get the token for creating a sales channel context.

tests/Controller/ExportControllerTest.php
TheKeymaster: The sales channel context should not be explicitly set. The controller must be responsible to get it himself based on the shopkey.

tests/Controller/ExportControllerTest.php
TheKeymaster: Please see [this comment](https://github.com/findologic/plugin-shopware-6/pull/131#discussion_r529667561).

tests/Export/SalesChannelServiceTest.php
TheKeymaster: Ok.

src/Findologic/Config/FinSearchConfigCollection.php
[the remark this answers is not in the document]
TheKeymaster: I see. I checked the Shopware code and they literally do it the same. Fine for me then.

src/Controller/FindologicConfigController.php
```diff
@@ -0,0 +1,84 @@
...
+     * @Route("/api/v{version}/_action/finsearch", name="api.action.finsearch.save", methods={"POST"})
+     */
+    public function saveConfiguration(Request $request): JsonResponse
+    {
+        $salesChannelId = $request->query->get('salesChannelId');
+        $languageId = $request->query->get('languageId');
+        $kvs = $request->request->all();
+        $this->saveKeyValues($salesChannelId, $languageId, $kvs);
+
+        return new JsonResponse(null, Response::HTTP_NO_CONTENT);
```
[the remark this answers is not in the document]
zaifastafa: `NoContentResponse` does not work. `The Response content must be a string or object implementing __toString(), "array" given.` The original implementation works.

src/Controller/FindologicConfigController.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        $languageId = $request->query->get('languageId');
+        $salesChannelId = $request->query->get('salesChannelId');
+
+        $values = $this->findologicConfigService->getDomain($salesChannelId, $languageId);
+        if (empty($values)) {
+            $json = '{}';
+        } else {
+            $json = json_encode($values, JSON_PRESERVE_ZERO_FRACTION);
+        }
```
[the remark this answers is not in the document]
zaifastafa: `NoContentResponse` does not work. `The Response content must be a string or object implementing __toString(), "array" given.` The original implementation works.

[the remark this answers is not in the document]
TheKeymaster: Your third argument explicitly tells Symfony that you want to parse JSON. If you remove it, it will work.

src/Controller/FindologicConfigController.php
```diff
@@ -0,0 +1,84 @@
...
+     * @Route("/api/v{version}/_action/finsearch", name="api.action.finsearch.save", methods={"POST"})
+     */
+    public function saveConfiguration(Request $request): NoContentResponse
+    {
+        $salesChannelId = $request->query->get('salesChannelId');
+        $languageId = $request->query->get('languageId');
+        $configs = $request->request->all();
+        $this->saveKeyValues($salesChannelId, $languageId, $configs);
+
+        // return new JsonResponse(null, Response::HTTP_NO_CONTENT);
```
TheKeymaster, line 61: Commented code, can be removed.

src/FinSearch.php
```diff
@@ -49,20 +51,28 @@ public function uninstall(UninstallContext $uninstallContext): void
             }
         }
 
-        parent::uninstall($uninstallContext);
+        if ($uninstallContext->keepUserData()) {
```
TheKeymaster, line 54: :ok_hand:

src/Findologic/Config/FindologicConfigService.php
TheKeymaster: Did you find the related build?

src/Migration/Migration1605714035FinSearchConfig.php
```diff
@@ -0,0 +1,104 @@
...
+            $data['sales_channel_id'] = Uuid::fromHexToBytes($salesChannelId);
+            try {
+                $connection->insert('finsearch_config', $data);
+            } catch (DBALException $ignored) {
+                // Do nothing here as configuration already exists
+            }
+        }
+
+        // After migrating, delete previous configuration as it won't be used anymore
+        $this->deleteOldConfiguration($connection);
```
TheKeymaster, line 89: Let's not remove the old configuration, as we still may need it, when we re-introduce some parts of "all sales channels".

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,301 @@
...
+     *         (
+     *             [loginRegistration] => Array
+     *                 (
+     *                     [showPhoneNumberField] => 'someValue'
+     *                 )
+     *         )
+     * )
+     * ```
+     */
+    private function parseFindologicConfiguration(FinSearchConfigCollection $collection): array
```
TheKeymaster, line 272: Isn't this parsing any configuration? If not the example is more confusing than it helps, if it is then rename the method.

src/Resources/app/administration/src/module/findologic-module/snippet/en-GB.json
```diff
@@ -7,7 +7,9 @@
     "fieldRequired": "Shopkey is required.",
     "invalidShopkey": "Invalid Shopkey.",
     "notRegisteredShopkey": "The given shopkey is not associated to any known service.",
+    "selectSalesChannel": "Please select a Sales Channel... ",
```
TheKeymaster, line 10: There seems to be a space at the end, please remove it.

tests/Controller/ExportControllerTest.php
```diff
@@ -180,4 +181,51 @@ protected function parsePluginVersion(): string
 
         return ltrim($parsed['version'], 'v');
     }
+
+    public function testCorrectTranslatedProductIsExported(): void
```
TheKeymaster, line 185: Now that test is pretty neat!

src/Resources/app/administration/src/module/findologic-module/components/findologic-config/findologic-config.html.twig
```diff
@@ -1,217 +1,106 @@
...
+        <sw-card-section>
+            <sw-card class="sw-card--grid"
+                     :title="$tc('findologic.settingForm.config.title')">
+                <sw-container>
+                    {% block findologic_credentials_card_container %}
+                        {% block findologic_credentials_settings %}
+                            <div class="findologic-settings-credentials-fields">
+                                <sw-text-field
```
TheKeymaster, line 10: Seems like all previous issues regarding the configuration page have been solved, just one minor thing is, that when removing the shopkey, I can save the configuration properly, but it still tells me that it is required. This should not be the case.

image

TheKeymaster (changes requested):

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,301 @@
...
+
+use function array_key_exists;
+use function array_shift;
+use function explode;
+use function gettype;
+use function is_array;
+
+class FindologicConfigService
+{
+    private const CONFIG_DOMAIN = 'FinSearch.';
```
TheKeymaster, line 30: Unused constant.

src/Findologic/Config/FindologicConfigService.php
zaifastafa: Technically yes, but in our service it will only parse Findologic config.

TheKeymaster: Then please rename it to just `parseConfiguration`, since we are already in the `Findologic` context, since this service is already called `FindologicConfigService`.

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -0,0 +1,307 @@
...
+    public function get(string $key, string $salesChannelId, string $languageId)
+    {
+        $config = $this->load($salesChannelId, $languageId);
+
+        $parts = explode('.', $key);
+
+        $pointer = $config;
+
+        foreach ($parts as $part) {
+            if (!\is_array($pointer)) {
```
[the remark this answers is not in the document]
zaifastafa: It is already part of PSR-12 ruleset `PSR12.Files.ImportStatement`

src/FinSearch.php
```diff
@@ -82,7 +95,7 @@ private function loadServiceXml($container, string $filePath): void
  *
  * @see https://github.com/shopware-blog/shopware-fastbill-connector/blob/development/src/FastBillConnector.php#L47
  */
-$loader = require_once __DIR__ . '/../vendor/autoload.php';
+$loader = require_once __DIR__.'/../vendor/autoload.php';
```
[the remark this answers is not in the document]
zaifastafa: I found this `Squiz.Strings.ConcatenationSpacing` but this actually does the opposite. It gives error if there is a space around the concat operator.

src/Findologic/Config/FindologicConfigService.php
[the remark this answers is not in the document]
TheKeymaster: How did this pass linting then?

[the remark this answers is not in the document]
zaifastafa: Might be of less severity?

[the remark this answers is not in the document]
TheKeymaster: It seems to be commented-out by the default ruleset:

https://github.com/squizlabs/PHP_CodeSniffer/blob/cda358face567f4cd8356c8d5411e38f30e2d3e9/src/Standards/PSR12/ruleset.xml#L101

Please add it explicitly to our `phpcs.xml`.

src/FinSearch.php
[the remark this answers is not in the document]
TheKeymaster: I just fiddled around, try this:

```xml
    <rule ref="Squiz.Strings.ConcatenationSpacing">
        <properties>
            <property name="spacing" value="1" />
        </properties>
    </rule>
```

src/Controller/FindologicConfigController.php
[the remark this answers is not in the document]
zaifastafa: No it does not seem to be working still. Shopware does it the same way as well in their controller.

src/FinSearch.php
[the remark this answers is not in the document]
zaifastafa: I saw this after signing off, so ill check it tomorrow

src/Findologic/Config/FindologicConfigService.php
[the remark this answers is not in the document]
zaifastafa: Have added it in the PR: https://github.com/findologic/plugin-shopware-6/blob/0756a0c04e78bb5a1989f93f674af6cb4f7d31eb/phpcs.xml#L36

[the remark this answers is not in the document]
TheKeymaster: Please add this change in this story, since it is a review issue of this one.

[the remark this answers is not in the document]
TheKeymaster: Nevermind, you already did :)

src/Utils/Utils.php
```diff
@@ -98,16 +100,19 @@ public static function multiByteRawUrlEncode(string $string): string
...
-            . (isset($parsedUrl['user']) ? "{$parsedUrl['user']}" : '')
-            . (isset($parsedUrl['pass']) ? ":{$parsedUrl['pass']}" : '')
-            . (isset($parsedUrl['user']) ? '@' : '')
-            . (isset($parsedUrl['host']) ? "{$parsedUrl['host']}" : '')
-            . (isset($parsedUrl['port']) ? ":{$parsedUrl['port']}" : '')
-            . (isset($parsedUrl['path']) ? "{$parsedUrl['path']}" : '')
-            . (isset($parsedUrl['query']) ? "?{$parsedUrl['query']}" : '')
-            . (isset($parsedUrl['fragment']) ? "#{$parsedUrl['fragment']}" : '');
+        return sprintf(
+            '%s%s%s%s%s%s%s%s%s%s',
```
TheKeymaster, line 104: That looks kinda funny :laughing:.

TheKeymaster (approved):

src/Controller/FindologicConfigController.php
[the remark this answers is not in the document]
TheKeymaster: Okay than let's stick with it. We still can change that in the future if we wanted to.

src/Findologic/Config/FindologicConfigService.php
TheKeymaster: I will check this myself. Not relevant right now.

PR #133, opened 2020-12-01
SW-543_fix_translated_non_seo_urls

TheKeymaster:


tests/Controller/ExportControllerTest.php
```diff
@@ -228,4 +214,189 @@ public function testCorrectTranslatedProductIsExported(): void
...
+                            $salesChannelContext->getSalesChannel()->getDomains()->first()->getSnippetSetId(),
+                        'url' => 'http://cool-url.com/german'
+                    ]
+                ]
+            ],
+        ], Context::createDefaultContext());
+
+        $this->enableFindologicPlugin(
+            $this->getContainer(),
+            self::VALID_SHOPKEY,
```
dblaichinger, line 263: It would be great if there is also a UK channel like described in the comment
> https://some-shop.de/findologic?shopkey=from-united-kingdom => https://shome-shop.co.uk/detail/...

Then I would go with calling http://cool-url.com/german?shopkey=from-united-kingdom and assert english urls are exported, not the /german one.

tests/Controller/ExportControllerTest.php
```diff
@@ -228,4 +214,189 @@ public function testCorrectTranslatedProductIsExported(): void
...
+     * within a different Sales Channel. E.g.
+     *
+     * Sales Channel "Germany" can be accessed with URL https://some-shop.de, therefore the configured export
+     * URL is set to it. Now there is a second domain for "United Kingdom", which can be accessed with the
+     * URL https://some-shop.co.uk. This test ensures that the products will use the UK domain, when calling
+     * the URL of the German Sales Channel with the shopkey of the United Kingdom Sales Channel:
+     * https://some-shop.de/findologic?shopkey=from-united-kingdom => https://shome-shop.co.uk/detail/...
+     *
+     * Runs in a separate process to not interfere with other tests.
+     * @runInSeparateProcess
```
dblaichinger, line 230: Why is this necessary? Because deletion of seo urls?

tests/Controller/ExportControllerTest.php
```diff
@@ -228,4 +214,189 @@ public function testCorrectTranslatedProductIsExported(): void
...
+
+        $this->assertSame(200, $response->getStatusCode());
+        $this->assertSame('text/xml; charset=UTF-8', $response->headers->get('content-type'));
+        $parsedResponse = new SimpleXMLElement($response->getContent());
+
+        $this->assertSame(1, (int)$parsedResponse->items->attributes()->count);
+        $this->assertSame($product->getId(), $parsedResponse->items->item->attributes()->id->__toString());
+        $this->assertSame(
+            'http://localhost/german/detail/' . $product->getId(),
+            $parsedResponse->items->item->urls->url->__toString()
```
dblaichinger, line 291: Isn't `__toString()` called by default?

tests/Controller/ExportControllerTest.php
```diff
@@ -228,4 +214,189 @@ public function testCorrectTranslatedProductIsExported(): void
...
+        $composerJsonContents = file_get_contents(__DIR__ . '/../../composer.json');
+        $parsed = json_decode($composerJsonContents, true);
+
+        // For release candidates, Shopware will internally format the version different, from how
+        // it is set in the composer.json.
+        return str_replace('rc.', 'RC', ltrim($parsed['version'], 'v'));
+    }
+
+    /**
+     * Unlike SalesChannelHelper::buildSalesChannelContext, which modifies the default sales channel, this
```
dblaichinger, line 319: Why not rename `buildSalesChannelContext` to `modifySalesChannelContext`?

tests/Controller/ExportControllerTest.php
```diff
@@ -228,4 +214,189 @@ public function testCorrectTranslatedProductIsExported(): void
...
+     * method creates an entirely new Sales Channel and returns an appropriate SalesChannelContext.
+     *
+     * @see SalesChannelHelper::buildSalesChannelContext
+     */
+    private function buildNewSalesChannelContext(
```
dblaichinger, line 324: `createSalesChannelContext`?

tests/Controller/ExportControllerTest.php
```diff
@@ -228,4 +214,189 @@ public function testCorrectTranslatedProductIsExported(): void
...
+        $languageId = array_values($languages->getElements())[1]->getId();
+
+        $currencyRepo = $this->getContainer()->get('currency.repository');
+        $currencies = $currencyRepo->search(
+            (new Criteria())->addFilter(new EqualsFilter('isoCode', 'EUR')),
+            Context::createDefaultContext()
+        );
+
+        $salesChannelContext = $this->buildNewSalesChannelContext($currencies, $languages);
+        $salesChannelContext->getSalesChannel()->setLanguageId($languageId);
```
dblaichinger, line 245: Doesn't `buildNewSalesChannelContext` already set the language id?

dblaichinger (changes requested):

tests/Controller/ExportControllerTest.php
TheKeymaster: No it is a SimpleXMLElement.

tests/Controller/ExportControllerTest.php
TheKeymaster: Yes, but this here is not a setting, but rather a context. With this method I tell Shopware that the user is using this language.

tests/Controller/ExportControllerTest.php
TheKeymaster: I couldn't pin down the exact reason, but the test runs fine when executed in a single scoped thread and I think that it may be related to Shopware, since they basically reset the database on each test and somehow that one product can then be not found. But then again, we are using a fake client to send the request. I think that even if it would pass (which it doesn't), going for the separate process makes this more stable overall.

tests/Controller/ExportControllerTest.php
TheKeymaster: This is just the other way around. The setup for this test is already pretty big. Also since I did test driven development, I know for a fact that this test is exactly testing what it should. The Shopware default Sales Channel is english, so we create a german one and call the english one with the german shopkey => Products are german. I do not see a real benefit of doing this the other way around.

tests/Controller/ExportControllerTest.php
TheKeymaster: That comment is not really accurate. It should be

... which **by default** modifies the default sales channel, this ...

We actually use it in this method to create an entirely new one, as `buildSalesChannelContext` uses `upsert`, which can either modify existing data, or if it doesn't exist, creates that data.

Updated the docblock.

tests/Controller/ExportControllerTest.php
dblaichinger: Ok understand, when there is an english channel by default, please somehow comment that its using the EN shopkey for the en here.

PR #137, opened 2020-12-09
SW-545_add_character_limit

TheKeymaster:


src/Struct/FindologicProduct.php
```diff
@@ -996,6 +997,10 @@ public function getCustomFields(): array
     protected function getCleanedAttributeValue($value)
     {
         if (is_string($value)) {
+            if (mb_strlen($value) > DataHelper::ATTRIBUTE_CHARACTER_LIMIT) {
+                return '';
+            }
+
```
dblaichinger, line 1003: > Please ensure this length validation is logged with product id and affected attribute.

Is libflexport handling logging?

dblaichinger (changes requested):

src/Struct/FindologicProduct.php
TheKeymaster: > Is libflexport handling logging?

No libflexport generally has nothing to do with logging.

I was trying to implement it, but it isn't trivial to allow the `FindologicProduct` to log stuff, without explicitly throwing an exception, which causes the product not to be exported. Preferably we introduce a logging mechanism in the future, so users of the extension plugin may also log some stuff.

dblaichinger: Added [SW-548](https://findologic.myjetbrains.com/youtrack/issue/SW-548) to improve logging.

dblaichinger (approved):

PR #138, opened 2020-12-09
SW-546_support_multiselect_custom_fields

TheKeymaster:


tests/Struct/FindologicProductTest.php
```diff
@@ -396,6 +396,38 @@ public function testProductWithCustomFields(): void
...
+            $this->getContainer(),
+            $this->salesChannelContext->getContext(),
+            $this->shopkey,
+            $customerGroupEntities,
+            new XMLItem('123')
+        );
+
+        $attributes = $findologicProduct->getCustomFields();
+        foreach ($attributes as $attribute) {
+            $this->assertEquals($attribute->getValues(), $productFields[$attribute->getKey()]);
```
dblaichinger, line 427: expected and actual should be switched.

src/Struct/FindologicProduct.php
```diff
@@ -991,10 +991,19 @@ public function getCustomFields(): array
...
     protected function getCleanedAttributeValue($value)
     {
+        if (is_array($value)) {
+            $values = [];
+            foreach ($value as $item) {
+                $values[] = $this->getCleanedAttributeValue($item);
+            }
+
+            return $values;
+        }
```
dblaichinger, line 1005: Should handling arrays still be in this method? Or should this logic be put before the call `getCleanedAttributeValue` in `getCustomFieldProperties`?

dblaichinger (changes requested):

src/Struct/FindologicProduct.php
TheKeymaster: I was thinking the same and I also had both solutions implemented once. I went with this, as this is simple and also allows customers who use the extension plugin to call this method and get their data, cleaned.

tests/Struct/FindologicProductTest.php
TheKeymaster: Oof. Then I might also do that for the other test, from where I copied that ^^.

src/Struct/FindologicProduct.php
```diff
@@ -991,10 +991,19 @@ public function getCustomFields(): array
 
     /**
      * @param string|int|bool $value
```
dblaichinger, line 993: Add `array<string, int>`

dblaichinger (changes requested):

dblaichinger (approved):

PR #141, opened 2020-12-17
SW-508 it must be possible to set configurations for all sales channels

zaifastafa:


src/Findologic/Config/FindologicConfigService.php
```diff
@@ -241,11 +262,22 @@ private function load(string $salesChannelId, string $languageId): array
     private function buildConfig(FinSearchConfigCollection $configs): array
```
TheKeymaster, line 262: I may have overseen this last review, but the method is called `buildConfig`, yet it requires `$configs`. I suggest going with the naming of `$options`, since a config contains multiple options. I am still fine with the naming `CONFIG_KEYS`, since it describes all keys inside of a config.

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -241,11 +262,22 @@ private function load(string $salesChannelId, string $languageId): array
...
+            $findologicConfig[$configKey] = null;
+        }
+
         foreach ($configs as $config) {
-            $keyExists = array_key_exists($config->getConfigurationKey(), $findologicConfig);
-            if (!$keyExists || !Utils::isEmpty($config->getConfigurationValue())) {
-                $findologicConfig[$config->getConfigurationKey()] = $config->getConfigurationValue();
+            $value = $config->getConfigurationValue();
+            $key = $config->getConfigurationKey();
+            $inheritedValuePresent = array_key_exists($key, $findologicConfig);
```
TheKeymaster, line 273: Isn't this always returning `false`? In the foreach above you explicitly set all values to `null`.

src/Migration/Migration1605714035FinSearchConfig.php
```diff
@@ -44,8 +50,8 @@ public function updateDestructive(Connection $connection): void
     }
 
     /**
-     * Custom Findologic configuration is implemented after v1.3.2, so we need to properly migrate any existing config
-     * from SystemConfig to FindologicConfig
+     * Custom Findologic configuration is implemented after 2.0, so we need to properly migrate any existing config
```
TheKeymaster, line 53:
```suggestion
     * Custom Findologic configuration is implemented in 2.0, so we need to properly migrate any existing config
```

src/Resources/app/administration/src/module/findologic-module/components/findologic-config/findologic-config.html.twig
```diff
@@ -7,58 +7,106 @@
...
+                                            :error="shopkeyErrorState"
+                                    >
+                                    </sw-text-field>
+                                </div>
+
+                                <sw-inherit-wrapper
                                         v-model="actualConfigData['FinSearch.config.active']"
-                                        :label="$tc('findologic.settingForm.config.active.label')"
-                                        :helpText="$tc('findologic.settingForm.config.active.tooltipText')"
+                                        :inheritedValue="selectedSalesChannelId == null ? null : allConfigs['null']['FinSearch.config.active']"
```
TheKeymaster, line 26: 1. Please be explicit by using `===`. If you are unsure if the variable is even set, you can also explicitly check for that as well by using `typeof`.
2. Move this logic to a helper method, since you do this check for basically all options.

src/Resources/app/administration/src/module/findologic-module/page/findologic-page/findologic-page.html.twig
```diff
@@ -84,7 +84,8 @@
                                             :allConfigs="allConfigs"
                                             :actualConfigData="actualConfigData"
                                             :shopkeyErrorState="shopkeyErrorState"
-                                            :selectedSalesChannelId="selectedSalesChannelId">
+                                            :selectedSalesChannelId="selectedSalesChannelId"
+                                            :selectedLanguageId="selectedLanguageId">
```
TheKeymaster, line 88: Why wasn't this required before?

tests/Findologic/Config/FindologicServiceTest.php
```diff
@@ -0,0 +1,77 @@
...
+        parent::setUp();
+
+        $this->findologicConfigService = new FindologicConfigService(
+            $this->getContainer()->get('finsearch_config.repository'),
+            $this->getContainer()->get(Connection::class),
+            $this->getContainer()->get('sales_channel.repository')
+        );
+    }
+
+    public function configurationProvider()
```
TheKeymaster, line 33: Add a typehint.

```suggestion
    public function configurationProvider(): array
```

tests/Findologic/Config/FindologicServiceTest.php
```diff
@@ -0,0 +1,77 @@
...
+            'Value is empty string' => [''],
+            'Value is string' => ['test'],
+            'Value is array' => [['foo' => 'bar']]
+        ];
+    }
+
+    /**
+     * @dataProvider configurationProvider
+     */
+    public function testConfigurationIsSetCorrectly($expectedValue)
```
TheKeymaster, line 51: I typically also add typehints for tests, but that one is up to you.

```suggestion
    public function testConfigurationIsSetCorrectly($expectedValue): void
```

tests/Findologic/Config/FindologicServiceTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $actual = $this->findologicConfigService->get('foo.bar', Defaults::SALES_CHANNEL);
+        static::assertSame($expectedValue, $actual);
+    }
+
+    /**
+     * @dataProvider configurationProvider
+     */
+    public function testInheritConfigurationIsReturned($expectedValue)
+    {
+        $this->findologicConfigService->set('foo.bar', $expectedValue);
```
TheKeymaster, line 73: Could you explicitly add `null` as last argument? This way the test would still test our case properly, even if we were to change the default.

TheKeymaster (changes requested):

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -241,11 +262,22 @@ private function load(string $salesChannelId, string $languageId): array
...
+            $isValueEmpty = !is_bool($value) && Utils::isEmpty($value);
```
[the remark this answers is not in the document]
zaifastafa: But a boolean false is not technically an empty value. So we might introduce some issues.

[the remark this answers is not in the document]
TheKeymaster: Ah I see, I misinterpreted this check. Why is a boolean empty in this case then? It isn't very obvious.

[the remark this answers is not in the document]
zaifastafa: Yeah I already fixed it in the last update.

src/Resources/app/administration/src/module/findologic-module/page/findologic-page/findologic-page.html.twig
TheKeymaster: Why was this resolved? There was no code change and my question hasn't been answered.

src/Findologic/Config/FindologicConfigService.php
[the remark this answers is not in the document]
TheKeymaster: What do you mean by that? The check `is_bool` doesn't even need to be here. Why was this check even introduced and is it required? Here I have a few examples:

`$value = true` => `!is_bool($value) is false` `Utils::isEmpty($value) is false` => `$isValueEmpty is false`
`$value = true` => `!is_bool($value) is false` `Utils::isEmpty($value) is false` => `$isValueEmpty is false`
`$value = "test"` => `!is_bool($value) is true` `Utils::isEmpty($value) is false` => `$isValueEmpty is false`
`$value = ["test"]` => `!is_bool($value) is true` `Utils::isEmpty($value) is false` => `$isValueEmpty is false`

The check `!is_bool` seems to me as it is not required in any case. I could be wrong but I am curious why it has been introduced in the first place, since it seems like it has no real impact on the returned result.

Please remove it if it turns out to really have no impact.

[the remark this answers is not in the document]
zaifastafa: Did you check my previous comment? I have already removed that check. It was how shopware was doing it so I figured it might have some significance but then in our case we did not have any importance so I have removed that check already.

src/Resources/app/administration/src/module/findologic-module/page/findologic-page/findologic-page.html.twig
zaifastafa: It was not required before because the implementation was done differently and it was checked directly in the module instead of the component.

PR #142, opened 2020-12-17
SW-551_fix_performance_for_image_filters

TheKeymaster:


src/Resources/views/storefront/component/listing/filter/filter-property-select.html.twig
```diff
@@ -2,5 +2,23 @@
...
     {% set displayType = element.displayType %}
-    {{ parent() }}
+
+    <span class="filter-property-select-preview"
+          title="{{ name }}"
+          {% if displayType == 'media' and media %}
+          style="background-color: {{ color }};background-image: url('{{ media }}')"
+          {% elseif displayType == 'color' and color %}
+          style="background-color: {{ color }};background-image:none"
+          {% endif %}
```
dblaichinger, line 12: Does that mean displayType = 'none' is the Shopware default image?

tests/Findologic/Response/Xml21ResponseParserTest.php
```diff
@@ -229,16 +229,19 @@ public function filterResponseProvider(): array
...
             (new ColorFilterValue('blau', 'blau', $color))
                 ->setColorHexCode('#3c6380')
                 ->setMedia(new Media('https://blubbergurken.io/farbfilter/blau.gif'))
+                ->setDisplayType('media')
         );
         $expectedColorFilter->addValue(
             (new ColorFilterValue('braun', 'braun', $color))
                 ->setColorHexCode('#94651e')
                 ->setMedia(new Media('https://blubbergurken.io/farbfilter/braun.gif'))
+                ->setDisplayType('media')
```
dblaichinger, line 244: Shouldn't there be tests for displayType color and none?

dblaichinger (changes requested):

src/Resources/views/storefront/component/listing/filter/filter-property-select.html.twig
TheKeymaster: Exactly. `none` will be used in case there is no URL and no image configured.

tests/Findologic/Response/Xml21ResponseParserTest.php
TheKeymaster: Definitely. Added one.

dblaichinger (approved):

PR #144, opened 2020-12-21
SW-557_fix_sales_frequency_memory

TheKeymaster:


src/Struct/FindologicProduct.php
```diff
@@ -488,9 +488,9 @@ protected function setSalesFrequency(): void
 
         /** @var EntityRepository $orderLineItemRepository */
         $orderLineItemRepository = $this->container->get('order_line_item.repository');
-        $orders = $orderLineItemRepository->search($criteria, $this->salesChannelContext->getContext());
+        $orders = $orderLineItemRepository->searchIds($criteria, $this->salesChannelContext->getContext());
 
-        $this->salesFrequency = $orders->count();
+        $this->salesFrequency = $orders->getTotal();
```
dblaichinger, line 493: Regarding unit test, would it be possible to mock orderLineItemRepository and ensure `searchIds` and afterwards check that `getTotal` is calle on the mocked order?

dblaichinger (changes requested):

src/Struct/FindologicProduct.php
TheKeymaster: That would also be possible, in fact I also thought about this. The only problem is that with such a solution, we wouldn't really ensure that our export gracefully handles orders in a memory-sensitive manner. We just ensure that specific methods are being called. I have mixed feeling about this.

dblaichinger: Sure this would be the better approach, but we shouldn't invest a lot of time in that. Ensuring that we don't go back to `search()` is probably good enough.

dblaichinger (changes requested):

src/Struct/FindologicProduct.php
TheKeymaster: You're right. Better than not testing this at all. I will add a test.

dblaichinger (approved):

PR #147, opened 2021-01-01
SW-498-export-list-prices-as-property

zaifastafa:


src/Struct/FindologicProduct.php
```diff
@@ -628,6 +628,18 @@ protected function setProperties(): void
             $value = $this->product->getManufacturer()->getMedia()->getUrl();
             $this->addProperty('vendorlogo', $value);
         }
+
+        if ($this->product->getPrice() && $this->product->getPrice()->first()) {
+            /** @var ProductPrice $price */
+            $price = $this->product->getPrice()->first();
```
TheKeymaster, line 634: Is getting the first price safe to use? In which case can there be multiple prices and can we catch such a case as well?

tests/Struct/FindologicProductTest.php
```diff
@@ -1210,4 +1212,44 @@ private function translateBooleanValue(bool $value): string
...
+        $findologicProduct = $findologicProductFactory->buildInstance(
+            $productEntity,
+            $this->router,
+            $this->getContainer(),
+            $this->shopkey,
+            [],
+            new XMLItem('123')
+        );
+
+        foreach ($findologicProduct->getProperties() as $property) {
```
TheKeymaster, line 1246: Does this test fail in case both `old_price` and `old_price_net` are not set?

TheKeymaster (changes requested):

src/Struct/FindologicProduct.php
zaifastafa: In general yes. Multiple prices are related to when there are prices for multiple currencies, so then each price would have a listing price accordingly (if any).

tests/Struct/FindologicProductTest.php
zaifastafa: No I dont think so. Will make the necessary changes.

src/Struct/FindologicProduct.php
TheKeymaster: Since we're using the Context of the Sales Channel when searching all products via DAL, does Shopware automatically give us only the price for the relevant Sales Channel? If not we need to fetch only those, which are actually relevant for the current Sales Channel. I want to avoid mixed up prices if customers provide multiple currency prices for different Sales Channels.

TheKeymaster (approved):

PR #154, opened 2021-01-19
SW-574_fix_filters_6_3_5_0 (for 2.0)

TheKeymaster:


tests/Findologic/Api/FindologicSearchServiceTest.php
```diff
@@ -170,4 +173,65 @@ public function testStagingQueryParameterWorksAsExpected(
...
+            function (string $name) use ($findologicService) {
+                if ($name === 'findologicService') {
+                    return $findologicService;
+                }
+
+                return null;
+            }
+        );
+
+        $request = new Request(['navigationId' => $this->getRootCategory()->getId()]);
```
dblaichinger, line 215: So root category is the unknown one, how does the FindologicService identify categories?

dblaichinger (changes requested):

tests/Findologic/Api/FindologicSearchServiceTest.php
TheKeymaster: The class `\FINDOLOGIC\FinSearch\Findologic\Request\Parser\NavigationCategoryParser` is responsible for finding out which category the current category is. The root category is considered unknown, as this is a category, which is not directly indexed by Findologic.

dblaichinger: Please add this as a comment, as it's not obvious.

dblaichinger (approved):

PR #157, opened 2021-01-22
Sw 521 disable filter values if they would cause a no results page

zaifastafa:


src/Findologic/Request/Handler/NavigationRequestHandler.php
```diff
@@ -106,7 +106,7 @@ public function handleRequest(ShopwareEvent $event): void
      * @throws ServiceNotAliveException
      * @throws UnknownCategoryException
      */
-    public function doRequest(ShopwareEvent $event, ?int $limit = null): Response
+    public function doRequest(ShopwareEvent $event, ?int $limit = null, ?bool $forceFilter = null): Response
```
TheKeymaster, line 109: Without deep knowledge of the implementation, you do not know what the parameter `$forceFilter` is supposed to do. Please add this in the method docblock. Same goes for `SearchNavigationRequestHandler::doRequest` and `SearchRequestHandler::doRequest`.

src/Findologic/Request/Handler/SearchRequestHandler.php
```diff
@@ -77,7 +77,7 @@ public function handleRequest(ShopwareEvent $event): void
      *
      * @throws ServiceNotAliveException
      */
-    public function doRequest(ShopwareEvent $event, ?int $limit = null): Response
+    public function doRequest(ShopwareEvent $event, ?int $limit = null, ?bool $forceFilter = null): Response
```
TheKeymaster, line 80: Do we even need `forceFilter` anyway? Seems like it has no usage here after all.

src/Findologic/Response/Xml21/Filter/Values/FilterValue.php
```diff
@@ -54,7 +54,7 @@ public function getTranslated(): TranslatedName
         return $this->translated;
     }
 
-    public function getUuid(): string
+    public function getUuid(): ?string
```
TheKeymaster, line 57: When could this be `null`?

src/Findologic/Api/FindologicSearchService.php
```diff
@@ -124,7 +126,7 @@ protected function handleRequest(
         $this->sortingService->handleRequest($event, $requestHandler);
     }
 
-    protected function allowRequest(ProductListingCriteriaEvent $event): bool
+    public function allowRequest(ProductListingCriteriaEvent $event): bool
```
TheKeymaster, line 129: It may be convenient to have this public, but I am against that. The `FindologicSearchService` should be aware of its state and must know itself if it should trigger or not. Make sure this method is called in `doFilter`. Every public method of this class should be usable, without having to think about any state.

In `SearchController::filter` you currently use this method to check if Findologic should be used or not. This shouldn't be strictly tied to this class. Instead find a similar approach like in `ProductListingRoute::load`:
```php
public function __construct()
{
    // ...
    $this->config = $config ?? new Config($systemConfigService, $serviceConfigResource);
}

public function load(
    string $categoryId,
    Request $request,
    SalesChannelContext $salesChannelContext
): ProductListingRouteResponse {
    $this->config->initializeBySalesChannel($salesChannelContext->getSalesChannel()->getId());
    $shouldHandleRequest = Utils::shouldHandleRequest(
        $request,
        $salesChannelContext->getContext(),
        $this->serviceConfigResource,
        $this->config,
        true
    );
    // ...
}
```

src/Findologic/Api/FindologicSearchService.php
```diff
@@ -213,4 +203,52 @@ protected function isCategoryPage(NavigationRequestHandler $handler, ProductList
 
         return !empty($isCategoryPage);
     }
+
+    public function doFilter(ProductListingCriteriaEvent $event): void
+    {
+        $limit = self::FILTER_REQUEST_LIMIT;
```
TheKeymaster, line 209: `$limit` is only used here. Instead use the constant and assign it directly to the method in `handleSelectableFilters`:
`$this->handleSelectableFilters($event, $handler, self::FILTER_REQUEST_LIMIT)`

src/Findologic/Api/FindologicSearchService.php
```diff
@@ -213,4 +203,52 @@ protected function isCategoryPage(NavigationRequestHandler $handler, ProductList
...
+        $filtersWithSmartSuggestBlocks = $this->parseFiltersFromResponse($response, $event);
+
+        $event->getCriteria()->addExtension('flAvailableFilters', $filtersWithSmartSuggestBlocks);
+    }
+
+    /**
+     * @param Response $response
+     * @param ProductListingCriteriaEvent $event
+     *
+     * @return mixed
```
TheKeymaster, line 234: Add a strict type hint and add it as method return type.

TheKeymaster: I also saw that this wasn't explicitly added to the `ResponseParser` class, therefore also add `FiltersExtension` as return type for this method in the response parser.

src/Utils/Utils.php
```diff
@@ -262,19 +285,50 @@ public static function fetchNavigationCategoryFromSalesChannel(
...
 
-    /**
-     * Builds the category path by removing the path of the parent (root) category of the sales channel.
-     * Since Findologic does not care about any root categories, we need to get the difference between the
-     * normal category path and the root category.
-     *
-     * @return string[]
-     */
-    private static function getCategoryBreadcrumb(array $categoryBreadcrumb, CategoryEntity $rootCategory): array
+    public static function parseFindologicFiltersForShopware(FiltersExtension $filtersExtension): array
```
TheKeymaster, line 288: This method shouldn't be in here. Please move it somewhere else. I think it fits best in `FilterHandler`.

tests/Storefront/Controller/SearchControllerTest.php
```diff
@@ -0,0 +1,102 @@
...
+            $this->getContainer(),
+            $mockFindologicService
+        );
+
+        $response = $controller->filter($request, $this->salesChannelContext);
+        $this->assertTrue($response->isOk());
+        $this->assertInstanceOf(JsonResponse::class, $response);
+
+        if ($active) {
+            $availableFilters = ['properties', 'rating', 'price', 'Farbe', 'Material', 'vendor', 'cat'];
```
TheKeymaster, line 94: Make sure that all filters do have their expected values in their id and their name (& translated name). Also ensure that the `properties` array contains the same data as the other filters.

src/Resources/app/storefront/src/js/filter-category-select.plugin.js
```diff
@@ -182,4 +204,63 @@ export default class FilterCategorySelectPlugin extends FilterBasePlugin {
     getSelected() {
         return DomAccess.querySelectorAll(this.el, `${this.options.checkboxSelector}:checked`, false);
     }
+
+    refreshDisabledState(filter) {
```
TheKeymaster, line 208: It looks like the category filter now is actually broken.

When I trigger an empty search, only filters on the first category level are active. Others are disabled:

image

As soon as I select a filter, all filters are being disabled:

image

The selected filters looks odd:

image

At the same time the filter isn't even selected and the URL looks like this: http://localhost:8000/search?cat=Outdoors%7COutdoors&order=score&p=1&search=

src/Resources/app/storefront/src/js/filter-property-select.plugin.js
```diff
@@ -0,0 +1,42 @@
+import FilterPropertySelectPlugin from 'src/plugin/listing/filter-property-select.plugin';
```
TheKeymaster, line 1: The size filter and color/vendor image is properly disabled:

image

image


But it doesn't seem to be the case for content or textile, so I am still able to get no results:

image

Please ensure this properly works.

tests/Storefront/Controller/SearchControllerTest.php
TheKeymaster: Also please add a test that ensures that if a filter doesn't have any more available values, their values are no longer inside the response.

TheKeymaster (changes requested):

src/Findologic/Request/Handler/SearchRequestHandler.php
zaifastafa: It is needed here because it is an abstract method and needs to match the signature of the abstract class. I will check if things work without it, and if so, will remove it altogether.

src/Findologic/Response/Xml21/Filter/Values/FilterValue.php
zaifastafa: As of now, it is null for categories filter.

src/Resources/app/storefront/src/js/filter-category-select.plugin.js
zaifastafa: @TheKeymaster 
> As soon as I select a filter, all filters are being disabled:

This was how it was originally implemented - and also was a requirement at that time, because we can only select one category filter. This is not related to the disabling filter change. 

> When I trigger an empty search, only filters on the first category level are active. Others are disabled:

This happens because we are only getting the main category from the response. Might need to change the response array so the children are also available in the array.

src/Resources/app/storefront/src/js/filter-property-select.plugin.js
zaifastafa: @TheKeymaster 
I had tested it and it was working. These all are the same filter type so we are not doing anything different for these. Can you verify the findologic API response. It happens that at times it sends all the filters back instead of only selected ones.

zaifastafa: I will have another look at it as well.

src/Resources/app/storefront/src/js/filter-category-select.plugin.js
TheKeymaster: As discussed, will be fixed in SW-578.

src/Resources/app/storefront/src/js/filter-property-select.plugin.js
TheKeymaster: As discussed, will be fixed in SW-578.

TheKeymaster:
Closed in favor of https://github.com/findologic/plugin-shopware-6/pull/165.

PR #158, opened 2021-01-28
SW-575_make_seourls_deletion_and_canonical_aware

TheKeymaster:


src/Struct/FindologicProduct.php
```diff
@@ -338,10 +337,18 @@ protected function getTranslatedSeoPath(): ?string
...
-        /** @var SeoUrlEntity|null $seoUrlEntity */
-        $seoUrlEntity = $this->getTranslatedEntity($seoUrlCollection);
+        $seoUrlEntities = $this->getTranslatedEntities($collectionWithoutDeletedEntities);
+        if (!$seoUrlEntities) {
+            return null;
+        }
+
+        $canonicalSeoUrl = $seoUrlEntities->filter(function (SeoUrlEntity $entity) {
+            return $entity->getIsCanonical();
+        })->first();
```
dblaichinger, line 349: Is `->first()` handling zero results after filter?

src/Struct/FindologicProduct.php
```diff
@@ -350,17 +357,15 @@ protected function getTranslatedDomainBaseUrl(): ?string
         $salesChannel = $this->salesChannelContext->getSalesChannel();
         $domainCollection = $salesChannel->getDomains();
 
-        /** @var SalesChannelDomainEntity|null $domainEntity */
-        $domainEntity = $this->getTranslatedEntity($domainCollection);
+        $domainEntities = $this->getTranslatedEntities($domainCollection);
 
-        return $domainEntity ? rtrim($domainEntity->getUrl(), '/') : null;
+        return $domainEntities->first() ? rtrim($domainEntities->first()->getUrl(), '/') : null;
```
dblaichinger, line 362: Is it also safe to call first() here?

dblaichinger (changes requested):

src/Struct/FindologicProduct.php
TheKeymaster: `first()` will return null, if it can not find an entitiy. In line 351 we would then fall back to the first SEO URL entity. And well, if that one doesn't exist as well, we do not have any SEO URL and therefore would fallback to the default product URL.

src/Struct/FindologicProduct.php
TheKeymaster: The same was done before just in a different place. If there is no domain, then we just return `null` here and similar to if we have no SEO URL, we fallback to the default product URL.

dblaichinger (approved):

PR #159, opened 2021-01-28
Sw 512 disable installation for incompatible shopware version

zaifastafa:


src/FinSearch.php
```diff
@@ -33,6 +40,34 @@ public function build(ContainerBuilder $container): void
...
+            // If the current shopware version does not come in between our lowest and highest supported version,
+            // we throw an exception and stop the plugin installation.
+            $isLower = version_compare($currentShopwareVersion, $lowestSupported, '<');
+            $isHigher = version_compare($currentShopwareVersion, $highestSupported, '>');
+            if ($isLower || $isHigher) {
+                $error = 'This plugin is not compatible with the used Shopware version';
+                throw new PluginNotCompatibleException($error);
+            }
+        }
+    }
```
TheKeymaster, line 69: Could this be abstracted to something along these lines:

```php
public function install(InstallContext $installContext): void
{
    // ...

    if (!$this->isCompatible()) {
        throw new PluginNotCompatibleException();
    }
}

private function isCompatible(): bool
{
    // ...
}
```

src/FinSearch.php
```diff
@@ -33,6 +40,34 @@ public function build(ContainerBuilder $container): void
...
+        if (!$parsed) {
+            return;
+        }
+
+        $requiredPackages = $parsed['require'];
+        if (isset($requiredPackages['shopware/core'])) {
+            $compatibleVersions = explode('||', $requiredPackages['shopware/core']);
+            $lowestSupported = ltrim(current($compatibleVersions), '^');
+            $highestSupported = ltrim(end($compatibleVersions), '^');
+            // If the current shopware version does not come in between our lowest and highest supported version,
```
TheKeymaster, line 60: This comment is probably not necessary after refactoring. Or as [Uncle Bob](https://twitter.com/unclebobmartin/status/870311898545258497) would say:

> A comment is a failure to express yourself in code.  If you fail, then write a comment; but try not to fail.

src/FinSearch.php
```diff
@@ -33,6 +40,34 @@ public function build(ContainerBuilder $container): void
...
+            // we throw an exception and stop the plugin installation.
+            $isLower = version_compare($currentShopwareVersion, $lowestSupported, '<');
+            $isHigher = version_compare($currentShopwareVersion, $highestSupported, '>');
```
TheKeymaster, line 63: Lets say `$highestSupported` is `6.3`, wouldn't this then fail for `6.3.5.0`?

TheKeymaster (changes requested):

src/Exceptions/PluginNotCompatibleException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace FINDOLOGIC\FinSearch\Exceptions;
+
+class PluginNotCompatibleException extends FindologicException
```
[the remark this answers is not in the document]
zaifastafa: Argh! Yeah I thought about doing that but then I checked other exception classes and we were not passing any messages in there so dont know why but I removed it 😆

src/FinSearch.php
zaifastafa: Very wise Uncle Bob! 👨🏼‍🎓

src/FinSearch.php
```diff
@@ -79,6 +97,32 @@ private function loadServiceXml($container, string $filePath): void
...
+        if (!$parsed) {
+            return true;
+        }
+
+        $requiredPackages = $parsed['require'];
+        if (isset($requiredPackages['shopware/core'])) {
+            $compatibleVersions = explode('||', $requiredPackages['shopware/core']);
+            $lowestSupported = current($compatibleVersions);
+            $highestSupported = end($compatibleVersions);
+            $isLower = Comparator::lessThan($currentShopwareVersion, $lowestSupported);
```
TheKeymaster, line 115: Fancy! I didn't knew that existed!

TheKeymaster (approved):

src/FinSearch.php
```diff
@@ -79,11 +99,52 @@ private function loadServiceXml($container, string $filePath): void
         $loader->load('services.xml');
     }
 
+    private function isCompatible(InstallContext $installContext): bool
```
TheKeymaster, line 102: Make this method public and static. This way we can properly unit-test it. Add the composer.json destination as optional parameter and override it in the tests. Ensure that

* composer.json contains `^6.1||^6.2||^6.3`
  * Shopware 6.1.0, 6.1.6, 6.2.0, 6.2.3, 6.3.0.0, 6.3.3.0 and 6.3.5.1 is compatible.
  * Shopware 6.4.0.0, 6.4.5.0, 6.5.1.2, 6.7.6.4 is not compatible.

This way we have this properly tested and can ensure that this works as expected.

src/FinSearch.php
```diff
@@ -79,11 +99,52 @@ private function loadServiceXml($container, string $filePath): void
...
+        return true;
+    }
+
     private function deleteFindologicConfig(): void
     {
         $connection = $this->container->get(Connection::class);
         $connection->executeUpdate('DROP TABLE IF EXISTS `finsearch_config`');
     }
+
+    private function isVersionLower(string $currentVersion, array $compatibleVersions): bool
```
TheKeymaster, line 131: These two methods need to be static as well, but please make them `protected`.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #161, opened 2021-02-03
Sw 561 use right dimension of product image in export

zaifastafa:


src/Struct/FindologicProduct.php
```diff
@@ -1129,4 +1145,21 @@ protected function getCatUrlPrefix(): string
 
         return rtrim($path, '/');
     }
+
+    protected function filterThumbnails(MediaThumbnailCollection $thumbnails): MediaThumbnailCollection
```
TheKeymaster, line 1149: If your method name is generic like this, either split their contents up into multiple smaller methods or add a docblock describing what actually is filtered. Also a method called `filterThumbnails` shouldn't be responsible for changing their order as well.

src/Struct/FindologicProduct.php
```diff
@@ -451,21 +455,33 @@ protected function setImages(): void
...
 
+            $filteredThumbnails = $this->filterThumbnails($thumbnails);
+
+            // Use the thumbnail as the main image if available, otherwise fallback to the directly assigned image.
+            $image = $filteredThumbnails->first() ?? $media;
+            if ($image) {
+                $encodedThumbnailUrl = $this->getEncodedUrl($image->getUrl());
+                if (!Utils::isEmpty($encodedThumbnailUrl)) {
+                    $this->images[] = new Image($encodedThumbnailUrl);
+                }
```
TheKeymaster, line 482: This exact block is duplicate to lines 467 - 470.

tests/Struct/FindologicProductTest.php
```diff
@@ -366,6 +372,375 @@ public function testProduct(): void
...
+                        ['width' => 200, 'height' => 200, 'highDpi' => false],
+                        ['width' => 400, 'height' => 400, 'highDpi' => false]
+                    ]
+                ]
+            ],
+            '0 thumbnails, the image directly assigned to the product is taken' => [
+                'thumbnails' => $data['cover']['media'] = ['thumbnails' => null]
+            ]
+        ];
+    }
```
TheKeymaster, line 408: Please add a newline.

tests/Struct/FindologicProductTest.php
```diff
@@ -366,6 +372,375 @@ public function testProduct(): void
...
+        );
+
+        $findologicImages = $findologicProduct->getImages();
+        $this->assertEquals($images, $findologicImages);
+    }
+
+    /**
+     * @return Image[]
+     */
+    private function getImages(ProductEntity $productEntity): array
```
TheKeymaster, line 438: Don't copy the actual implementation to ensure your code works properly. Either use static strings to compare your expected data or if this is not possible, since a random URL will be generated, try to get the data directly from the database.
In addition to that, the code in this method is way to complex for a test and would on is own already require a bunch of tests.

I know this method already existed before, but even before it was a bit on the edge and now it is just way too much.

tests/Struct/FindologicProductTest.php
```diff
@@ -1367,6 +1480,23 @@ public function testProductListPrice(?string $currencyId, bool $isPriceAvailable
         $this->assertSame($isPriceAvailable, $hasListPriceNet);
     }
 
+    private function filterThumbnails(MediaThumbnailCollection $thumbnails): MediaThumbnailCollection
```
TheKeymaster, line 1483: See my other comment. This is not how tests are done.

tests/Struct/FindologicProductTest.php
```diff
@@ -366,6 +372,375 @@ public function testProduct(): void
...
+        $encodedPath = array_map('\FINDOLOGIC\FinSearch\Utils\Utils::multiByteRawUrlEncode', $urlPath);
+        $parsedUrl['path'] = implode('/', $encodedPath);
+
+        return Utils::buildUrl($parsedUrl);
+    }
+
+    /**
+     * @return Attribute[]
+     */
+    private function getAttributes(ProductEntity $productEntity): array
```
TheKeymaster, line 511: Why are methods moved everywhere? Please move them back to where they were in the first place. It is not just hard to review but also not really beneficial to anyone.

tests/Struct/FindologicProductTest.php
TheKeymaster: A test should test that your code works, by knowing before the test starts, what the outcome is. Tests should prevent bugs in the implementation and here the implementation was just copied, which may result in bugs. Imagine you have this class:

```php
class Calculator {
    public function multiply(int $number1, int $number2): int
    {
        return $number1 - $number2;
    }
}
```

You would expect `multiply` to multiply these two numbers, but in this case it substracts one from another, which is a bug. But since you are copying the implementation, you aren't even aware of this bug:

```php
class CalculatorTest extends TestCase {
    public function testAddsTwoNumbersAddsThemTogether(): void
    {
        $calculator = new Calculator();
        $number1 = 3;
        $number2 = 2;

        $expectedResult $this->getMultiply($number1, $number2);
        $actualResult = $calculator->multiply($number1, $number2);

        
        $this->assertSame($expectedResult, $actualResult);
    }

    private function getMultiply(int $number1, int $number2): int
    {
        return $number1 - $number2;
    }
}
```

Although you wouldn't copy this, because it is obvious, it is less obvious when you copy more complex code. How would you know that the implementation doesn't have a bug when you test that the code you have copied from the implementation is the same? You can't without reasonable effort.
Please rethink this and make sure not to copy the actual implementation. We want tests to avoid bugs and not to tell us everything is fine, when it isn't the case.

TheKeymaster (changes requested):

src/Struct/FindologicProduct.php
zaifastafa: No actually, I had already tried refactoring this, but a ThumbnailMediaEntity is different than a MediaEntity so the typehint would not match. But I will try it refactoring another way.

tests/Struct/FindologicProductTest.php
zaifastafa: Actually the IDE does that to sort related methods together. I have disabled that setting. Won't happen anymore.

tests/Struct/FindologicProductTest.php
zaifastafa: Changing this method is unnecessary work in my opinion, as is it just filtering and sorting an array. If you are worried about the correctness of the implementation then that is justified - but we can add a simple assertion to make sure it filters how we expect it.

src/Struct/FindologicProduct.php
TheKeymaster: As a hint, if there isn't a general class, you may also just use `Entity` as a return type and add `ThumbnailMediaEntity|MediaEntity` to the return type in the docblock.

tests/Struct/FindologicProductTest.php
TheKeymaster: I would understand sorting, but filtering shouldn't be done here. In your test you must know your expectations before you start a test. If you just use the same logic, as in the implementation, you aren't really ensuring that the order or the filtering is right.

I would accept this helper if there was a separate test that ensures that the thumbnails are properly filtered and sorted, where you already know the outcome of the filtering and sorting.

tests/Struct/FindologicProductTest.php
TheKeymaster: Yes please. Reordering methods just makes it really hard to review.

src/Struct/FindologicProduct.php
```diff
@@ -1129,4 +1139,25 @@ protected function getCatUrlPrefix(): string
...
+        });
+
+        $filteredThumbnails->sort(function (MediaThumbnailEntity $a, MediaThumbnailEntity $b) {
+            return $a->getWidth() <=> $b->getWidth();
+        });
+
+        return $filteredThumbnails;
+    }
+
+    private function addMediaUrl(string $encodedUrl, bool $isThumbnail = false): void
```
TheKeymaster, line 1156: This may be split up into two methods. `setImageUrl` and `addThumbnailUrl`.

TheKeymaster (changes requested):

src/Struct/FindologicProduct.php
TheKeymaster: This wasn't addressed.

src/Struct/FindologicProduct.php
```diff
@@ -1129,4 +1137,36 @@ protected function getCatUrlPrefix(): string
...
+            return $a->getWidth() <=> $b->getWidth();
+        });
+
+        return $filteredThumbnails;
+    }
+
+    /**
+     * @param MediaThumbnailEntity|MediaEntity $media
+     */
+    private function setImageUrl(Entity $media): void
```
TheKeymaster, line 1157: Please change the visibility of this method to `protected`, otherwise plugin extensions can not call them and would need to define it on their own.

src/Struct/FindologicProduct.php
```diff
@@ -1129,4 +1137,36 @@ protected function getCatUrlPrefix(): string
...
+    {
+        $encodedUrl = $this->getEncodedUrl($media->getUrl());
+        if (!Utils::isEmpty($encodedUrl)) {
+            $this->images[] = new Image($encodedUrl);
+        }
+    }
+
+    private function addThumbnailUrl(MediaThumbnailEntity $media): void
```
TheKeymaster, line 1165: Please change the visibility of this method to `protected`, otherwise plugin extensions can not call them and would need to define it on their own.

tests/Struct/FindologicProductTest.php
```diff
@@ -1449,4 +1480,167 @@ public function testCatUrlsContainDomainPathAsPrefix(): void
...
+        });
+
+        $filteredThumbnails->sort(function (MediaThumbnailEntity $a, MediaThumbnailEntity $b) {
+            return $a->getWidth() <=> $b->getWidth();
+        });
+
+        return $filteredThumbnails;
+    }
+
+    public function widthSizesProvider()
```
TheKeymaster, line 1590: Missing return type.

tests/Struct/FindologicProductTest.php
```diff
@@ -1449,4 +1480,140 @@ public function testCatUrlsContainDomainPathAsPrefix(): void
...
+     * @dataProvider thumbnailProvider
+     */
+    public function testImageThumbnailsAreFilteredAndSortedByWidth(?array $thumbnails, ?int $expectedWidth): void
+    {
+        $productEntity = $this->createTestProduct(['cover' => ['media' => ['thumbnails' => $thumbnails]]]);
+        $mediaCollection = $productEntity->getMedia();
+        $media = $mediaCollection->getMedia();
+        $thumbnailCollection = $media->first()->getThumbnails();
+
+        $filteredThumbnails = $this->sortAndFilterThumbnailsByWidth($thumbnailCollection);
```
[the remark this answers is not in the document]
TheKeymaster: This wasn't addressed.

tests/Traits/DataHelpers/ProductHelper.php
```diff
@@ -5,23 +5,34 @@
...
+                    'visibility' => 20
+                ]
+            ]
+        ], $overrides));
+    }
+
+    public function createTestProduct(
+        array $data = [],
+        bool $withVariant = false,
+        bool $doRecursive = false
```
TheKeymaster, line 34: `$doRecursive` isn't really telling anything. Either rename the variable or add a description in the docblock.

TheKeymaster (changes requested):

tests/Struct/FindologicProductTest.php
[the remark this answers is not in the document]
zaifastafa: Yes it has. There is a new provider for the width tests.

TheKeymaster (approved):

PR #163, opened 2021-02-10
SW-580_fix_error_on_manual_listing_requests

TheKeymaster:


src/Findologic/Api/FindologicSearchService.php
```diff
@@ -213,4 +212,16 @@ protected function isCategoryPage(NavigationRequestHandler $handler, ProductList
...
+    protected function disableFindologicService(ProductListingCriteriaEvent $event): void
+    {
+        /** @var FindologicService|null $findologicService */
+        $findologicService = $event->getContext()->getExtension('findologicService');
+        if (!$findologicService) {
+            $findologicService = new FindologicService();
+            $event->getContext()->addExtension('findologicService', $findologicService);
+        }
+
+        $findologicService->disable();
```
dblaichinger, line 225: Why do we need to call disable when the findologicService doesn't exist? Wouldn't this already be "disabled"?

tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
```diff
@@ -1107,4 +1113,36 @@ public function testHandleResultDoesNotThrowExceptionWhenCalledManually(): void
...
+                if ($param === 'navigationId') {
+                    return Uuid::randomHex();
+                }
+
+                return null;
+            });
+
+        $subscriber = $this->getDefaultProductListingFeaturesSubscriber();
+        $subscriber->handleListingRequest($eventMock);
+    }
```
dblaichinger, line 1147: This test doesn't assert anything, is it possible to check if a findologicService has been set to the context?

dblaichinger (changes requested):

src/Findologic/Api/FindologicSearchService.php
TheKeymaster: In the usual case it will be gotten from the context. Creating a new instance is really a corner case, as a plugin explicitly would need to call this. At this point I do not really care where the `FindologicService` comes from, and in which state it currently is, I just want to disable it.

tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
TheKeymaster: Since `$eventMock` is a mock itself, and we mocked `getExtension`, to already return `null`, I ensured that `addExtension` is properly called with a new instance of `FindologicService`. To ensure this actually happened, I also declared `$addExtensionWasCalled`, which only is being set to `true` if `addExtension` is ever called with the argument `findologicService`. 

Please check out the code changes in this regard.

dblaichinger: Thanks, this makes it easier to follow. Also when someone accidentally breaks it, easier to fix!

src/Findologic/Api/FindologicSearchService.php
dblaichinger: > as a plugin explicitly would need to call this

Why does another plugin need to call this, seems wrong to me?!

Further we have several usages of `$event->getContext()->getExtension('findologicService')`, will those work fine without this check?

Screen Shot 2021-02-10 at 18 08 51

TheKeymaster: Issue here is the whole stack doesn't get executed (e.g. `FrontendSubscriber`). In order for your described code to actually work, there already must be an existing `FindologicService`, which must be active. Therefore disabling it when the request fails or the alivetest fails, is the proper way to deal with this.

PR #164, opened 2021-02-11
SW-579_export_single_size_thumbnail

TheKeymaster:


src/Struct/FindologicProduct.php
```diff
@@ -1169,4 +1167,21 @@ protected function addThumbnailUrl(MediaThumbnailEntity $media): void
             $this->images[] = new Image($encodedUrl, Image::TYPE_THUMBNAIL);
         }
     }
+
+    /**
+     * Goes through all given thumbnails, and will only add one thumbnail width as thumbnail. This avoids
+     * the adding of thumbnails in all various sizes.
+     */
```
dblaichinger, line 1174: "Go through all given thumbnails and only add one thumbnail image. This avoids exporting thumbnails in all various sizes."

src/Struct/FindologicProduct.php
```diff
@@ -1169,4 +1167,21 @@ protected function addThumbnailUrl(MediaThumbnailEntity $media): void
...
+    protected function addThumbnailImages(MediaThumbnailCollection $thumbnails): void
+    {
+        $imageIds = [];
+        foreach ($thumbnails as $thumbnailEntity) {
+            if (in_array($thumbnailEntity->getMediaId(), $imageIds)) {
+                continue;
+            }
+
+            $this->addThumbnailUrl($thumbnailEntity);
```
dblaichinger, line 1183: How are those sorted? Is the first one always the right one?

tests/Struct/FindologicProductTest.php
```diff
@@ -1542,19 +1556,77 @@ public function thumbnailProvider(): array
...
+                    [
+                        'width' => 1800,
+                        'height' => 1800,
+                        'highDpi' => false,
+                        'url' => 'https://via.placeholder.com/1800'
+                    ],
+                ],
+                'expectedImages' => [
+                    '800x800',
+                    '800x800',
```
dblaichinger, line 1612: So there is one actual image and one thumbnail, right? Can you make this more clear as they both have the same sizes.

dblaichinger (changes requested):

src/Struct/FindologicProduct.php
TheKeymaster: Thumbnails are sorted by width, therefore the first thumbnail is the thumbnail with the smallest width:

https://github.com/findologic/plugin-shopware-6/blob/e990ecdea5199c439ffa71db276c0e88cbd81e98/src/Struct/FindologicProduct.php#L473

dblaichinger (approved):

PR #165, opened 2021-02-16
SW-578_fix_current_issues_for_disabling_filters

zaifastafa:


src/Resources/app/storefront/src/scss/category-filter.scss
```diff
@@ -11,6 +11,11 @@
   .custom-checkbox {
     max-width: 290px;
 
+    &.disabled,
+    &:disabled {
+      opacity: calc(.3/.5);
```
TheKeymaster, line 16: This causes some really odd opacity, until the point comes where it is actually almost invisible:
image

Please remove this and let it handle like before.

src/Findologic/Request/Handler/FilterHandler.php
```diff
@@ -235,17 +236,19 @@ private function parseFindologicFiltersForShopware(FiltersExtension $filterExten
...
-                    if (!$filter instanceof CategoryFilter) {
+
+                if ($filter instanceof CategoryFilter) {
+                    $this->handleCategoryFilters($values, $filterValues);
+                } else {
+                    foreach ($values as $value) {
+                        $valueId = $value->getUuid() ?? $value->getId();
+                        $filterValues[] = [
+                            'id' => $valueId,
+                            'translated' => ['name' => $value->getTranslated()->getName()]
```
TheKeymaster, line 250: This was probably overseen, but the translated value should also contain `$valueId` here.

src/Findologic/Request/Handler/FilterHandler.php
```diff
@@ -268,4 +271,22 @@ private function parseFindologicFiltersForShopware(FiltersExtension $filterExten
 
         return array_merge($actualResult, $result);
     }
+
+    /**
+     * @param FilterValue[] $values
+     */
+    private function handleCategoryFilters(array $values, array &$filterValues): void
```
TheKeymaster, line 278: I really dislike passing values per reference, if it doesn't serve an immediate purpose, other than making it easier accessible. Try to return the value instead.

src/Findologic/Request/Handler/FilterHandler.php
```diff
@@ -268,4 +271,22 @@ private function parseFindologicFiltersForShopware(FiltersExtension $filterExten
 
         return array_merge($actualResult, $result);
     }
+
+    /**
+     * @param FilterValue[] $values
```
TheKeymaster, line 276: The second argument isn't properly type-hinted.

src/Storefront/Controller/SearchController.php
```diff
@@ -43,32 +43,18 @@ class SearchController extends StorefrontController
     /** @var FindologicSearchService */
     private $findologicSearchService;
 
-    /**
```
TheKeymaster, line 46: There are still a few imports in this class, which haven't been removed. Please remove them.

TheKeymaster (changes requested):

src/Resources/app/storefront/src/scss/category-filter.scss
zaifastafa: Disable class is applied differently for the category filter due to nested values. Which is why it adds opacity for deeper nested links. If you want to just disable the filter, that I can do and it will use the default styling instead of using opacity. But then the disable styling for category will be different than other filters.

src/Findologic/Request/Handler/FilterHandler.php
zaifastafa: This would require using `array_merge` inside the loop (due to recursive nature of the method), which is not very efficient in my opinion.

src/Resources/app/storefront/src/scss/category-filter.scss
```diff
@@ -39,4 +39,8 @@
 .filter-multi-select-cat > div {
   overflow-x: hidden;
   overflow-y: auto;
-}
\ No newline at end of file
+}
+
+.fl-disabled {
```
TheKeymaster, line 44: This fix solves the multiple opacities, but now also active filters and filters that are still available will also have this opacity:

image

src/Findologic/Request/Handler/FilterHandler.php
TheKeymaster: Sure makes sense then.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #174, opened 2021-03-12
Sw 581 improve export performance sales frequency

zaifastafa:


src/Struct/FindologicProduct.php
```diff
@@ -495,8 +499,18 @@ public function getSalesFrequency(): int
 
     protected function setSalesFrequency(): void
     {
+        $lastMonthDate = new DateTimeImmutable();
```
TheKeymaster, line 502: Isn't it easier to just do `new DateTimeImmutable('-1 month');`?

tests/Traits/DataHelpers/OrderHelper.php
```diff
@@ -0,0 +1,96 @@
...
+            'deliveries' => [],
+            'context' => '{}',
+            'payload' => '{}',
+        ];
+
+        $data = array_merge($data, $overrideData);
+        $this->getContainer()->get('order.repository')->create([$data], Context::createDefaultContext());
+    }
+
+    private function fetchFirstIdFromTable(string $table): string
```
TheKeymaster, line 90: This method isn't strictly tied to an "order helper".

tests/Struct/FindologicProductTest.php
```diff
@@ -1811,4 +1828,47 @@ public function testProductPromotionIsExported(?bool $markAsTopSeller, string $e
         $this->assertNotEmpty($values);
         $this->assertSame($expected, current($values));
     }
+
+    public function salesFrequencyProvider()
```
TheKeymaster, line 1832: Missing return type.

TheKeymaster (changes requested):

PR #176, opened 2021-03-17
SW-583_shopkey_validations

zaifastafa:


src/Resources/app/administration/src/module/findologic-module/page/findologic-page/index.js
```diff
@@ -56,7 +56,13 @@ Component.register('findologic-page', {
             this.shopkeyErrorState = null;
             if (this.isValidShopkey) {
                 this._isStagingRequest();
+            } else {
```
TheKeymaster, line 59:
```js
// ...
else {
    if {
       // ...
    }
}
```

Also known as `elseif:

```suggestion
            } else if (/** ... */) {
```

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #179, opened 2021-03-30
SW-497_add_support_for_shopware_6_4

TheKeymaster:
ToDo list:
* [x] Plugin can be installed and activated.
* [x] Plugin configuration can be changed.
* [x] Export properly works.
* [x] Search/Navigation with filters work.
* [x] Merchandising Features work.
* [x] Find fallback for admin configuration for versions lower than 6.4.
* [x] Check if we need to override sw-extension-card-bought as well.
* [x] Find a way to change the design for the new configuration.
* [x] FINAL: Add Shopware 6.4 to the test matrix and fix all tests.


.github/workflows/phpunit.yml
```diff
@@ -82,7 +73,7 @@ jobs:
           ./psh.phar init-test-databases
           ./psh.phar cache
 
-      # TODO: Re-enable the deprecation helper in SW-497.
+      # Deprecation Helper disabled until https://github.com/shopware/platform/pull/1759 is merged.
```
dblaichinger, line 76: Isn't it merged now? 👍

phpunit.legacy.xml.dist
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- Legacy testing configuration for Shopware versions < 6.4 and < PHPUnit 9 -->
```
dblaichinger, line 3:
```suggestion
<!-- Legacy test configuration for Shopware < 6.4 and PHPUnit < 9 -->
```

src/Resources/app/administration/src/module/findologic-module/page/findologic-page/index.js
```diff
@@ -145,6 +150,30 @@ Component.register('findologic-page', {
...
+         */
+        updateHeaderBarBackLink() {
+            let foundModule = null;
+            Shopware.Module.getModuleRegistry().forEach((module) => {
+                if (foundModule) {
+                    return;
+                }
+
+                if (module.manifest.title === 'sw-extension-store.title') {
+                    foundModule = module;
```
dblaichinger, line 167: Return here when foundModule is set?

src/Utils/Utils.php
```diff
@@ -216,6 +228,21 @@ public static function isFindologicEnabled(SalesChannelContext $context): bool
         return $findologicService ? $findologicService->getEnabled() : false;
     }
 
+    public static function disableFindologicWhenEnabled(SalesChannelContext $context): void
```
dblaichinger, line 231: Why is this relevant now and not in previous SW versions?

tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
```diff
@@ -574,15 +580,19 @@ private function initMocks(): void
             $sorting->setKey('score');
             $sorting->setFields(['score' => ['field' => '_score', 'order' => 'asc']]);
             $sorting->setUniqueIdentifier('score');
-            $productSorting = new ProductSortingCollection([$sorting]);
-            $this->productListingSortingRegistry->expects($this->any())
-                ->method('getProductSortingEntities')
-                ->willReturn($productSorting);
+            $this->sortingCollection = new ProductSortingCollection([$sorting]);
+
+            if (method_exists($this->productListingSortingRegistry, 'getProductSortingEntities')) {
```
dblaichinger, line 585: Shouldn't the version check ensure that the method does exist?

dblaichinger (changes requested):

.github/workflows/phpunit.yml
TheKeymaster: Yeah well, they merged it into `trunk`, which means that the comment should actually reflect `... when 6.4 is available`.

src/Resources/app/administration/src/module/findologic-module/page/findologic-page/index.js
TheKeymaster: You can't return inside a `.forEach` in JavaScript. That's also how Shopware solves this in their own code.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#description

> There is no way to stop or break a forEach() loop other than by throwing an exception.

src/Utils/Utils.php
TheKeymaster: This caused filter on the initial page (which is only enabled by default on your local development environment), to not be clickable anymore. It also happened for previous Shopware versions, but this is pretty rare case, as it only happens when Findologic is enabled on category pages as well.

tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
TheKeymaster: I prefer not do do a version check if I can do it without one. A simple `method_exists` is sufficient here.

dblaichinger (changes requested):

tests/Core/Content/Product/SalesChannel/Listing/ProductListingFeaturesSubscriberTest.php
TheKeymaster: This version check is not really related.

* The class `ProductListingSortingRegistry` exists since Shopware 6.3.2.
* The method `ProductListingSortingRegistry::getProductSortingEntities` was removed with Shopware 6.4.

dblaichinger (approved):

PR #187, opened 2021-05-25
SW-600-support-usage-of-pushattrib

zaifastafa:


src/Findologic/Request/Handler/FilterHandler.php
```diff
@@ -23,6 +23,7 @@ class FilterHandler
     protected const FILTER_DELIMITER = '|';
     protected const MIN_PREFIX = 'min-';
     protected const MAX_PREFIX = 'max-';
+    protected static $ignoreList = ['pushAttrib'];
```
TheKeymaster, line 26: This can also be a `const`, we do not want to manipulate this during runtime.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #188, opened 2021-05-27
SW-606-range-slider-issue-fix

zaifastafa:


src/Resources/app/storefront/package.json
```diff
@@ -1,19 +0,0 @@
-{
```
TheKeymaster, line 1: I think the `README.md` still contains information about running `npm install`, can we remove this?

src/Resources/app/storefront/src/js/filter-slider-range.plugin.js
```diff
@@ -318,4 +315,18 @@ export default class FilterSliderRange extends FilterBasePlugin {
       this.slider.noUiSlider.set([null, this._inputMax.value]);
     }
   }
+
+  refreshDisabledState(filter) {
```
TheKeymaster, line 319: Is this method called by the parent? I can not see any caller in this class.

src/Resources/app/storefront/src/js/filter-slider-range.plugin.js
```diff
@@ -318,4 +315,18 @@ export default class FilterSliderRange extends FilterBasePlugin {
...
+    const properties = filter['price'];
```
TheKeymaster, line 320: What happens for other range-slider filters, if you always explicitly select the `price`?

src/Resources/app/storefront/src/js/filter-slider-range.plugin.js
```diff
@@ -49,7 +46,7 @@ export default class FilterSliderRange extends FilterBasePlugin {
       step: this.options.price.step,
       range: {
         'min': this.options.price.min,
-        'max': this.options.price.max,
+        'max': this.options.price.max === this.options.price.min ? this.options.price.min + 1 : this.options.price.max
```
TheKeymaster, line 49: Maybe move this to a separate method?

TheKeymaster: Does it break the range-slider when min and max are the same?

TheKeymaster (changes requested):

src/Resources/app/storefront/src/js/filter-slider-range.plugin.js
zaifastafa: Yes it is.

src/Resources/app/storefront/src/js/filter-slider-range.plugin.js
zaifastafa: Yes it gives an error that min/max cannot be same.

TheKeymaster (approved):

PR #189, opened 2021-06-07
SW-616_do_not_override_components_if_fl_is_inactive

TheKeymaster:


.github/workflows/phpunit.yml
```diff
@@ -70,4 +70,4 @@ jobs:
       - name: Run test suite
         run: |
           cd development/custom/plugins/plugin-shopware-6
-          composer test
+          SYMFONY_DEPRECATIONS_HELPER=disabled composer test
```
TheKeymaster, line 73: Shopware internally has some deprecations, which are way too many that we can get rid of, so we need to disable the deprecation helper again, unfortunately.

src/Resources/views/storefront/component/listing/filter/filter-multi-select.html.twig
```diff
@@ -1,16 +1,20 @@
 {% sw_extends '@Storefront/storefront/component/listing/filter/filter-multi-select.html.twig' %}
 
 {% block component_filter_multi_select %}
-    <div class="{% if hidden %}d-none {% endif %}filter-multi-select filter-multi-select-{{ name }} filter-panel-item{% if not sidebar %} dropdown{% endif %}"
-         data-{{ pluginSelector }}="true"
-         data-{{ pluginSelector }}-options='{{ dataPluginSelectorOptions|json_encode }}'>
+    {% if not context.context.extensions.findologicService.enabled %}
```
dblaichinger, line 4: Does enabled stand for navigation?

tests/Struct/FindologicProductTest.php
```diff
@@ -369,9 +368,13 @@ public function testProduct(): void
         );
 
         $salesChannel = $this->salesChannelContext->getSalesChannel();
-        $domain = $salesChannel->getDomains()->first()->getUrl();
+        $domains = $salesChannel->getDomains();
```
dblaichinger, line 371: Is there already a `default.headless` included?

dblaichinger (commented):
Two questions

src/Resources/views/storefront/component/listing/filter/filter-multi-select.html.twig
TheKeymaster: The `FindologicService` is aware of the state and tells us if Findologic should be used or not.

tests/Struct/FindologicProductTest.php
TheKeymaster: Yes that's the Headless Sales Channel, which is always created by default. Only recently with Shopware 6.4 it appears to also to include an URL which is not really an URL, so we just filter all of these headless Sales Channels out of this.

PR #193, opened 2021-06-23
SW-582 Add separate route for dynamic product groups

zaifastafa:


src/Controller/ExportController.php
```diff
@@ -95,9 +103,43 @@ public function export(Request $request, ?SalesChannelContext $context): Respons
...
+        $excludeProductGroups = filter_var($request->get('excludeProductGroups'), FILTER_VALIDATE_BOOLEAN);
+        $dynamicProductGroupService = $this->getDynamicProductGroupService();
+        if (!$excludeProductGroups && !$dynamicProductGroupService->isDynamicProductGroupWarmedUp()) {
+            $errorMessage = 'Dynamic Product Groups have not been warmed up yet. ';
+            $errorMessage .= 'This may cause missing categories! ';
+            $errorMessage .= "Warm them up by calling the route '/findologic/dynamic-product-groups', ";
+            $errorMessage .= 'or disable fetching of Dynamic Product Groups by adding the query parameter ';
+            $errorMessage .= "'excludeProductGroups=true'";
+
+            return new JsonResponse(['error' => $errorMessage], Response::HTTP_PRECONDITION_REQUIRED);
```
TheKeymaster, line 115: I would suggest we create a separate class for this response, as the contents here bloat the actual code and distracts from the real implementation. E.g.

```php
class PreconditionFailedResponse extends JsonResponse
{
    public const CODE_DYNAMIC_PRODUCT_GROUPS_NOT_WARMED_UP = 0;

    public function __construct(int $errorCode, /* ... */)
    {
        switch ($errorCode) {
             case self::CODE_DYNAMIC_PRODUCT_GROUPS_NOT_WARMED_UP:
                 $errorMessage = '...';
                 break;
             default:
                 $errorMessage = 'Unknown';
        }
        
        parent::__construct(/* ... */);
    }
}
```

I suggest adding it inside the `Export` folder and create a new folder called `Responses`

src/Controller/ExportController.php
```diff
@@ -68,6 +73,9 @@ class ExportController extends AbstractController
     /** @var SalesChannelService|null */
     private $salesChannelService;
 
+    /** @var int */
+    private $total = 0;
```
TheKeymaster, line 77: This should not be a property inside this class. Handle it as a simple variable inside the method it is actually needed and pass it, in case a method requires it.

src/Controller/ExportController.php
TheKeymaster: As this is also considered an "error response", please make sure that `$this->validate()` is responsible for it.

src/Controller/ExportController.php
```diff
@@ -95,9 +103,43 @@ public function export(Request $request, ?SalesChannelContext $context): Respons
...
     }
 
+    /**
+     * @Route("/findologic/dynamic-product-groups", name="frontend.findologic.export.productgroup",
+     *     options={"seo"="false"}, methods={"GET"})
+     */
+    public function exportProductGroup(Request $request, ?SalesChannelContext $context): Response
+    {
+        $this->initialize($request, $context);
+        if ($errorResponse = $this->validate()) {
```
TheKeymaster, line 128: When adding the validation for the dynamic product groups, you may want to add another method, that does simply that but without validating the set dynamic product groups. E.g. `$this->validateWithoutDynamicProductGroups()`. Feel free to change this name, as this name isn't very creative.

src/Export/DynamicProductGroupService.php
```diff
@@ -121,28 +132,37 @@ public function setSalesChannel(SalesChannelEntity $salesChannelEntity): void
...
             return;
         }
-        $cacheItem->set(serialize($products));
+
+        // Set the total dynamic groups in cache once during warm up
+        $totalCacheItem = $this->getDynamicGroupsTotalCacheItem();
+        if ($totalCacheItem && !$totalCacheItem->isHit()) {
+            $total = $this->getTotalDynamicProductGroupsCount();
+            $this->setTotalInCache($totalCacheItem, $total);
+        }
```
TheKeymaster, line 147: Can this be moved to a separate method

src/Export/DynamicProductGroupService.php
```diff
@@ -151,16 +171,18 @@ public function isWarmedUp(): bool
     private function parseProductGroups(): array
     {
         $criteria = $this->buildCriteria();
+        $criteria->setOffset($this->start)->setLimit($this->count);
 
         /** @var CategoryCollection $categories */
         $categories = $this->categoryRepository->search($criteria, $this->context)->getEntities();
 
-        if ($categories === null || empty($categories->getElements())) {
+        if ($categories === null || !$categories->count()) {
```
TheKeymaster, line 179: I prefer being explicit over than implicit, by exactly checking `$categories->count() === 0`

src/Export/DynamicProductGroupService.php
```diff
@@ -218,10 +246,109 @@ private function buildCriteria(): Criteria
...
+     * Gets the total count of dynamic product groups from cache
+     */
+    public function getDynamicProductGroupTotalFromCache(): int
+    {
+        $cacheItem = $this->getDynamicGroupsTotalCacheItem();
+        if ($cacheItem && $cacheItem->isHit()) {
+            $cacheItem->expiresAfter(self::CACHE_LIFETIME_PRODUCT_GROUP);
+            $this->cache->save($cacheItem);
+
+            return (int)$cacheItem->get();
```
TheKeymaster, line 280: When is this not an int? Can the cache not handle numbers?

src/Export/DynamicProductGroupService.php
```diff
@@ -218,10 +246,109 @@ private function buildCriteria(): Criteria
...
+        if ($categories === null || !$categories->count()) {
+            return null;
+        }
+
+        return $categories;
+    }
+
+    public function getDynamicProductGroupWarmedUpCacheItem(): CacheItemInterface
+    {
+        $id = sprintf('%s_%s_dpg_warmup', self::CACHE_ID_PRODUCT_GROUP, $this->shopkey);
```
TheKeymaster, line 336: Don't use abbreviations. Someone not familiar with `dpg` would never know what it means.

src/Export/DynamicProductGroupService.php
```diff
@@ -218,10 +246,109 @@ private function buildCriteria(): Criteria
         return $criteria;
     }
 
-    private function getCacheItem(): CacheItemInterface
+    private function getExportPaginationCacheItem(): CacheItemInterface
```
TheKeymaster, line 249: The name export pagination cache item, is a bit confusing. Can this be renamed so its clear what it is used for?

src/Export/DynamicProductGroupService.php
```diff
@@ -218,10 +246,109 @@ private function buildCriteria(): Criteria
...
+    }
+
+    public function getDynamicProductGroupWarmedUpCacheItem(): CacheItemInterface
+    {
+        $id = sprintf('%s_%s_dpg_warmup', self::CACHE_ID_PRODUCT_GROUP, $this->shopkey);
+
+        return $this->cache->getItem($id);
+    }
+
+    public function isDynamicProductGroupWarmedUp(): bool
```
TheKeymaster, line 341: Good idea of moving this as a responsibility of the service, but the reason we didn't do this is because that stuff like the `ProductStreamBuilder` may not be available for older Shopware versions. Have you ensured that this still works properly?

src/Export/DynamicProductGroupService.php
```diff
@@ -218,10 +246,109 @@ private function buildCriteria(): Criteria
...
 
-    private function getCacheItem(): CacheItemInterface
+    private function getExportPaginationCacheItem(): CacheItemInterface
+    {
+        $id = sprintf('%s_%s_%s', self::CACHE_ID_PRODUCT_GROUP, $this->shopkey, $this->start);
+
+        return $this->cache->getItem($id);
+    }
+
+    private function setTotalInCache(CacheItemInterface $cacheItem, $total): void
```
TheKeymaster, line 256: As this whole thing now is quite complex, can we think about restructuring the logic in here a bit? We have like 12 methods here that all do something, but without knowing the exact details, I do not really know how they're connected.

src/Export/DynamicProductGroupService.php
```diff
@@ -218,10 +246,109 @@ private function buildCriteria(): Criteria
...
+
+            return (int)$cacheItem->get();
+        }
+
+        return 0;
+    }
+
+    /**
+     * If we have reached the last page of the dynamic product group export, we set a flag in cache to
+     * know that the dynamic product groups are warmed up
```
TheKeymaster, line 288: Good idea of adding a comment for this. Please try to always add `.` to the end of your sentences. There are many comments where this applies, please have a look, I only add this one comment but it applies to all newly added comments.

src/Export/DynamicProductGroupService.php
```diff
@@ -218,10 +246,109 @@ private function buildCriteria(): Criteria
...
+                continue;
+            }
+
+            $total++;
+        }
+
+        return $total;
+    }
+
+    private function getCategoriesWithProductGroups(Criteria $criteria): ?CategoryCollection
```
TheKeymaster, line 322: This method is clearly lying about its implementation. It is not getting categories with product groups, but actually getting categories, based on the given criteria and if there are no categories, null is returned.

src/Controller/ExportController.php
```diff
@@ -95,9 +103,43 @@ public function export(Request $request, ?SalesChannelContext $context): Respons
             return $errorResponse;
         }
 
+        $excludeProductGroups = filter_var($request->get('excludeProductGroups'), FILTER_VALIDATE_BOOLEAN);
```
TheKeymaster, line 106: You can also do `$request->query->getBoolean()`.

TheKeymaster (changes requested):

src/Controller/ExportController.php
zaifastafa: I think these specific type methods are not available for older version.

src/Export/DynamicProductGroupService.php
zaifastafa: Yes because we are passing the warm up flag as true if the version is lower than 6.3 so the warm up logic is not called to trigger the product stream builder.

src/Export/DynamicProductGroupService.php
zaifastafa: Yes it can, but the return value is a `mixed` value, so explicitly typecasting it into an int

src/Controller/ExportController.php
TheKeymaster: No this is related to the Symfony version, and the `getBoolean` method already exists since Symfony 2.6.
symfony_vers

* https://symfony.com/doc/2.8/components/http_foundation.html
* https://symfony.com/doc/current/components/http_foundation.html

src/Export/DynamicProductGroupService.php
```diff
@@ -121,46 +124,49 @@ public function setSalesChannel(SalesChannelEntity $salesChannelEntity): void
...
 
-            return true;
-        }
+    public function isDynamicProductGroupsCached(): bool
+    {
+        return $this->cacheHandler->isDynamicProductGroupsCached();
+    }
 
-        return false;
+    public function getDynamicProductGroupTotal()
```
TheKeymaster, line 150: What is the purpose of adding a method that will only call a single other method? This bloats the class, which already does so many things even more.

src/Export/CacheHandler.php
```diff
@@ -0,0 +1,157 @@
...
+            $cacheItem->expiresAfter(self::CACHE_LIFETIME_PRODUCT_GROUP);
+            $this->cache->save($cacheItem);
+
+            return true;
+        }
+
+        return false;
+    }
+
+    public function getCachedCategoryIds(int $offset): array
```
TheKeymaster, line 142: Some methods require an offset, while others don't and there is no clear distinction between why. Make the purpose of the method more clear, by giving it a more explicit name.

src/Export/CacheHandler.php
```diff
@@ -0,0 +1,157 @@
...
+    {
+        $id = sprintf('%s_%s_total', self::CACHE_ID_PRODUCT_GROUP, $this->shopkey);
+
+        return $this->cache->getItem($id);
+    }
+
+    /**
+     * Gets the total count of dynamic product groups from cache.
+     */
+    private function getDynamicProductGroupTotalFromCache(): int
```
TheKeymaster, line 78: This method does not need a docblock. The purpose is already clear by the name of the method.

src/Export/CacheHandler.php
```diff
@@ -0,0 +1,157 @@
...
+    public function setShopkey(string $shopkey): void
+    {
+        $this->shopkey = $shopkey;
+    }
+
+    /**
+     * If we have reached the last page of the dynamic product group export, we set a flag in cache to
+     * know that the dynamic product groups are warmed up.
+     */
+    public function dynamicProductGroupWarmUp(int $start, int $count, int $total = 0): void
```
TheKeymaster, line 35: A method name should always start with a verb.

```suggestion
    public function warmUpDynamicProductGroups(int $start, int $count, int $total = 0): void
```

src/Export/CacheHandler.php
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public function setShopkey(string $shopkey): void
+    {
+        $this->shopkey = $shopkey;
+    }
+
+    /**
+     * If we have reached the last page of the dynamic product group export, we set a flag in cache to
+     * know that the dynamic product groups are warmed up.
```
TheKeymaster, line 33: This docblock describes a detail, which is only clear when inside the method. Either add this docblock to the method, which is responsible for doing this, or add it above the code that actually implements this functionality.

src/Export/CacheHandler.php
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private function getDynamicProductGroupOffsetCacheItem(int $offset): CacheItemInterface
+    {
+        $id = sprintf('%s_%s_%s', self::CACHE_ID_PRODUCT_GROUP, $this->shopkey, $offset);
+
+        return $this->cache->getItem($id);
+    }
+
+    private function setTotalInCache(CacheItemInterface $cacheItem, $total): void
```
TheKeymaster, line 61: `$total` is missing a typehint.

src/Export/CacheHandler.php
```diff
@@ -0,0 +1,157 @@
...
+        return $totalCacheItem && $totalCacheItem->isHit();
+    }
+
+    public function setDynamicProductGroupTotal(int $total): void
+    {
+        $totalCacheItem = $this->getDynamicGroupsTotalCacheItem();
+        $this->setTotalInCache($totalCacheItem, $total);
+    }
+
+    public function setDynamicProductGroupOffset($products, int $offset): void
```
TheKeymaster, line 104: `$products` is missing a typehint.

src/Export/CacheHandler.php
```diff
@@ -0,0 +1,157 @@
...
+            $cacheItem->expiresAfter(self::CACHE_LIFETIME_PRODUCT_GROUP);
+            $this->cache->save($cacheItem);
+
+            return true;
+        }
+
+        return false;
+    }
+
+    public function isDynamicProductGroupsCached(): bool
```
TheKeymaster, line 128: public and private methods are all floating around places. I usually prefer having public methods before private methods, in case the class has so many responsibilities as this class. This way the next developer don't has to scroll around and search the method he or she needs. Having such a distinction also helps getting an overall overview of what the class is responsible for.

TheKeymaster:
```suggestion
    public function areDynamicProductGroupsCached(): bool
```

src/Export/CacheHandler.php
```diff
@@ -0,0 +1,157 @@
...
+            $cacheItem->expiresAfter(self::CACHE_LIFETIME_PRODUCT_GROUP);
+            $this->cache->save($cacheItem);
+
+            return (int)$cacheItem->get();
+        }
+
+        return 0;
+    }
+
+    public function isDynamicProductGroupTotalCached(): bool
```
TheKeymaster, line 91: Rename all occurrences of `DynamicProductGroup` to `DynamicProductGroups`. This way we are consistent with the naming of the methods.

src/Export/CacheHandler.php
```diff
@@ -0,0 +1,157 @@
...
+class CacheHandler
+{
+    private const CACHE_ID_PRODUCT_GROUP = 'fl_product_groups';
+    private const CACHE_LIFETIME_PRODUCT_GROUP = 60 * 11;
+
+    /** @var CacheItemPoolInterface */
+    protected $cache;
+
+    /** @var string */
+    protected $shopkey;
```
TheKeymaster, line 19: Initially the shopkey is `null`, which should be reflected by the docblock.

src/Export/CacheHandler.php
```diff
@@ -0,0 +1,157 @@
...
+    {
+        $cacheItem = $this->getDynamicGroupsTotalCacheItem();
+        if ($cacheItem && $cacheItem->isHit()) {
+            $cacheItem->expiresAfter(self::CACHE_LIFETIME_PRODUCT_GROUP);
+            $this->cache->save($cacheItem);
+
+            return (int)$cacheItem->get();
+        }
+
+        return 0;
```
TheKeymaster, line 88: Does this even make sense? Shouldn't we throw an exception instead?

src/Export/DynamicProductGroupService.php
```diff
@@ -76,37 +74,42 @@ class DynamicProductGroupService
 
     private function __construct(
         ContainerInterface $container,
-        CacheItemPoolInterface $cache,
+        CacheHandler $cacheHandler,
```
TheKeymaster, line 77: There are many backwards compatible breaks within your changes, which means we would need a major release. Is there any way to reduce the manual effort when upgrading the plugin and having an extension that explicitly overrides this or other classes, which received backwards compatibility breaks?

src/Export/DynamicProductGroupService.php
TheKeymaster: Please remove these single-line methods and simply call the `cacheHandler` methods directly.

src/Controller/ExportController.php
```diff
@@ -94,10 +100,35 @@ public function export(Request $request, ?SalesChannelContext $context): Respons
...
         }
+        if ($errorResponse = $this->validateDynamicGroupPrecondition($request)) {
+            return $errorResponse;
+        }
 
         return $this->doExport();
     }
 
+    /**
+     * @Route("/findologic/dynamic-product-groups", name="frontend.findologic.export.productgroup",
```
TheKeymaster, line 111:
```suggestion
     * @Route("/findologic/dynamic-product-groups", name="frontend.findologic.export.dynamic_product_groups",
```

tests/Export/DynamicProductGroupServiceTest.php
```diff
@@ -60,11 +60,26 @@ class DynamicProductGroupServiceTest extends TestCase
...
+    /**
+     * @var string
+     */
+    protected $cacheTotalKey;
+
+    /**
+     * @var string
+     */
+    private $cacheWarmupKey;
+
```
TheKeymaster, line 82: Change these class properties to private helper methods instead.

TheKeymaster (changes requested):

src/Export/DynamicProductGroupService.php
TheKeymaster: How can this be `mixed`? Will it return a string or why are we casting this to an int?

src/Export/DynamicProductGroupService.php
TheKeymaster: :+1:

src/Export/DynamicProductGroupService.php
zaifastafa: This is done because the export controller is using the dynamic product service for a reason and the cache handling is being done in that service. These one-line methods are being called from the export controller and the export controller does not use the cache handler directly.

src/Export/DynamicProductGroupService.php
zaifastafa: will have to check the possibilities.

src/Export/DynamicProductGroupService.php
```diff
@@ -121,46 +124,49 @@ public function setSalesChannel(SalesChannelEntity $salesChannelEntity): void
...
-        $this->cache->save($cacheItem);
+
+        $this->cacheDynamicProductGroupTotal();
+        $this->cacheDynamicProductOffset($products);
+
+        $this->cacheHandler->dynamicProductGroupWarmUp($this->start, $this->count);
     }
 
-    public function isWarmedUp(): bool
+    public function isOffsetWarmedUp(): bool
```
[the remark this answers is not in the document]
TheKeymaster: I saw that the method name now is `isCurrentOffsetWarmedUp`. I understand that something else probably doesn't really make sense, as this method really checks this, so it is fine.

src/Export/DynamicProductGroupService.php
```diff
@@ -20,93 +18,78 @@
...
+    /** @var DynamicProductGroupCacheHandler */
+    protected $cacheHandler;
+
+    /** @var SalesChannelEntity */
+    protected $salesChannel;
+
+    /** @var EntityRepositoryInterface */
+    protected $categoryRepository;
 
     private function __construct(
```
TheKeymaster, line 55: When we're already at it, please change this constructor to be `public`. Currently it is a PITA overriding this service, so we should at least make it easier for extensions to override this.

src/Export/DynamicProductGroupCacheHandler.php
```diff
@@ -0,0 +1,155 @@
...
+        return $totalCacheItem && $totalCacheItem->isHit();
+    }
+
+    public function setDynamicProductGroupTotal(int $total): void
+    {
+        $totalCacheItem = $this->getDynamicGroupsTotalCacheItem();
+        $this->setTotalInCache($totalCacheItem, $total);
+    }
+
+    public function setDynamicProductGroupsOffset(array $products, int $offset): void
```
TheKeymaster, line 102: Add a docblock typehint for the structure of `array $products`.

src/Export/CacheHandler.php
TheKeymaster: True but this was in relation to an already existing class. As this is a new class, please feel free to properly move the methods. I also don't expect many more review issues anyway at this point.

tests/Export/DynamicProductGroupServiceTest.php
```diff
@@ -304,4 +298,33 @@ private function createProducts(): array
 
         return $products;
     }
+
+    private function getProductStreamCategories(Context $context): array
```
TheKeymaster, line 302: Add a docblock of the return type, so the IDE already knows what data is returned here.

TheKeymaster (changes requested):

TheKeymaster (approved):

src/Export/DynamicProductGroupService.php
```diff
@@ -24,46 +27,36 @@ class DynamicProductGroupService
 {
     public const CONTAINER_ID = 'fin_search.dynamic_product_group';
 
-    protected ProductStreamBuilderInterface $productStreamBuilder;
-
     protected EntityRepositoryInterface $productRepository;
 
-    protected Context $context;
+    private EntityRepositoryInterface $categoryRepository;
```
msilinskas: Why this variable has private visability?

Also, the order of local variables is not the same like in the construct

src/Export/DynamicProductGroupService.php
```diff
@@ -225,37 +232,26 @@ protected function getTotalDynamicProductGroupsCount(): int
...
-    /**
-     * Sets the dynamic product groups in cache for each pagination. This is required so that each
-     * subsequent export request fetches the correct dynamic product groups for that offset.
-     */
-    protected function cacheDynamicProductOffset(array $products): void
+    protected function isLastPage(): bool
     {
-        $this->cacheHandler->setDynamicProductGroupsOffset($products, $this->start);
+        $currentTotal = $this->exportConfig->getStart() + $this->exportConfig->getCount();
+        return $currentTotal >= $this->getDynamicProductGroupsTotal();
```
msilinskas, line 254: blank line before return statement?

msilinskas (changes requested):

msilinskas (approved):

PR #196, opened 2021-07-14
Sw 595 direct integration cat urls must be generated recursively

zaifastafa:


src/Struct/FindologicProduct.php
```diff
@@ -1036,55 +967,27 @@ protected function parseCategoryAttributes(
...
                 $this->navigationCategory
             );
 
             if (!Utils::isEmpty($categoryPath)) {
-                $categories[] = $categoryPath;
+                $categories = array_merge($categories, [$categoryPath]);
+            }
+
+            // We only export "cat_urls", and "cat" recursively if there is Direct Integration, otherwise we only
+            // export "cat" without recursion.
```
TheKeymaster, line 980:
```suggestion
            // Only export `cat_url`s recursively if integration type is Direct Integration.
            // Note that this also applies for the `cat` attribute.
```

src/Struct/FindologicProduct.php
```diff
@@ -1036,55 +967,27 @@ protected function parseCategoryAttributes(
...
+            if ($this->isDirectIntegration()) {
+                $catUrls = array_merge($catUrls, $this->urlBuilderService->buildCatUrls($categoryEntity));
+                $parentCategory = explode('_', $categoryPath);
+                $categories = array_merge($categories, $parentCategory);
             }
         }
     }
 
-    protected function getCatUrlPrefix(): string
+    private function isDirectIntegration(): bool
```
TheKeymaster, line 989: Never add `private` methods in here, otherwise plugin extensions have no way of overriding or calling them.

TheKeymaster: Please check if there is any way to tell our linter to disallow private methods in a specific class.

src/Utils/Utils.php
```diff
@@ -387,4 +386,40 @@ public static function filterSalesChannelDomainsWithoutHeadlessDomain(
...
+     */
+    public static function flat(array $array): array
+    {
+        $flattened = [];
+        array_walk_recursive($array, static function ($a) use (&$flattened) {
+            $flattened[] = $a;
+        });
+
+        return $flattened;
+    }
```
TheKeymaster, line 424: This should no longer be needed after refactoring.

tests/Struct/FindologicProductTest.php
```diff
@@ -333,20 +348,29 @@ public function testProductPrices(?Price $price, ?string $exception): void
         }
     }
 
+    public function directIntegrationProvider()
```
TheKeymaster, line 351: 1. Rename this to `integrationTypeProvider`
2. Refactor this test so the data provider returns the integration type e.g. `API` or `Direct Integration` (don't use the constants, use the direct values).

tests/Struct/FindologicProductTest.php
```diff
@@ -333,20 +348,29 @@ public function testProductPrices(?Price $price, ?string $exception): void
...
+    {
+        return [
+            'Direct Integration' => [true],
+            'API' => [false],
```
TheKeymaster, line 355:
```suggestion
            'Integration type is Direct Integration' => [
                'integrationType' => 'Direct Integration'
            ],
            'Integration type is API' => [
                'integrationType' => 'API'
            ],
```

TheKeymaster: Please add pre-defined results instead of dynamically getting the required result. This may be done by creating a separate test with a data provider. This is what the test cases in the story description were designed for E.g.

```php
public function categoryAndCatUrlWithIntegrationTypeProvider(): array
{
    return [
        'Integration type is API and category is at first level' => [
             'integrationType' => 'API',
             'categories' => [
                 'id-of-assigned-category'
             ],
             'expectedCategories' => [
                'SomeCategoryName'
             ],
             'expectedCatUrls' => [],
        ],
        'Integration type is API with nested categories' => [
             'integrationType' => 'API',
             'categories' => [
                 'id-of-assigned-category'
             ],
             'expectedCategories' => [
                'Some_Category_Path' // Not recursive
             ],
             'expectedCatUrls' => [],
        ],
        'Integration type is DI and category is at first level' => [
             'integrationType' => 'Direct Integration',
             'categories' => [
                 'id-of-assigned-category'
             ],
             'expectedCategories' => [
                'SomeCategoryName'
             ],
             'expectedCatUrls' => [
                'SomeCategoryPath'
             ],
        ],
        'Integration type is DI with nested categories' => [
             'integrationType' => 'Direct Integration',
             'categories' => [
                 'id-of-assigned-category'
             ],
             'expectedCategories' => [
                'Some_Category_Path',
                'Some_Category',
                'Some',
             ],
             'expectedCatUrls' => [
                'Some/Category/Path',
                'Some/Category',
                'Some',
             ],
        ],
        'Integration type is unknown and category is at first level' => [
             'integrationType' => null,
             'categories' => [
                 'id-of-assigned-category'
             ],
             'expectedCategories' => [
                'SomeCategoryName'
             ],
             'expectedCatUrls' => [
                'SomeCategoryPath'
             ],
        ],
        'Integration type is unknown with nested categories' => [
             'integrationType' => null,
             'categories' => [
                 'id-of-assigned-category'
             ],
             'expectedCategories' => [
                'Some_Category_Path',
                'Some_Category',
                'Some',
             ],
             'expectedCatUrls' => [
                'Some/Category/Path',
                'Some/Category',
                'Some',
             ],
        ],
    ];
}

public function integrationTypeSpecialCharacterProvider(): array
{
    return [
        'Integration type is API' => [
            'integrationType' => 'API',
            'attributeValue' => '<><<<##test',
            'expectedAttributeValue' => 'test',
        ],
        'Integration type is DI' => [
            'integrationType' => 'Direct Integration',
            'attributeValue' => '<><<<##test',
            'expectedAttributeValue' => '<><<<##test',
        ],
        'Integration type is unknown' => [
            'integrationType' => 'Direct Integration',
            'attributeValue' => '<><<<##test',
            'expectedAttributeValue' => 'test',
        ],
    ];
}
```

TheKeymaster (changes requested):

src/Export/UrlBuilderService.php
```diff
@@ -0,0 +1,322 @@
...
+            return null;
+        }
+
+        $criteria = new Criteria([$parentId]);
+        $criteria->addAssociation('seoUrls');
+
+        $result = $this->categoryRepository->search($criteria, $this->salesChannelContext->getContext());
+
+        return $result->first();
+    }
```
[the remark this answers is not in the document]
TheKeymaster: See https://github.com/findologic/plugin-shopware-6/pull/196#discussion_r676489703

tests/Struct/FindologicProductTest.php
TheKeymaster: See https://github.com/findologic/plugin-shopware-6/pull/196#discussion_r676517958

src/Struct/FindologicProduct.php
zaifastafa: No there is not. There is one that checks if visibility is there or not but I didnt find anything to define specific visibility for a specific class

src/Struct/FindologicProduct.php
```diff
@@ -1000,4 +981,24 @@ protected function isApiIntegration(): bool
...
+        $parentCategory = explode('_', $categoryPath);
+
+        return array_merge($categories, $parentCategory);
+    }
+
+    /**
+     * For API Integrations, we have to remove special characters from the attribute key as a requirement for
+     * sending data via API.
+     */
+    private function processAttributeKey(?string $key): ?string
```
TheKeymaster, line 996: * Calling something `process` isn't really telling anything, and it also does not matter what "it" really processes. How about `getAttributeKey`. The internals do not really matter for anyone, except the one really interested in what it does and then the method name helps understanding.
* Change the visibility from `private` to `protected`, as otherwise extensions can not call it.

src/Struct/FindologicProduct.php
```diff
@@ -136,7 +136,8 @@ public function __construct(
         ContainerInterface $container,
         string $shopkey,
         array $customerGroups,
-        Item $item
+        Item $item,
+        ?Config $config = null
```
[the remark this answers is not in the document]
TheKeymaster: This comment is still open, without any code change.

src/Struct/FindologicProduct.php
```diff
@@ -829,11 +777,13 @@ protected function setCategoriesAndCatUrls(): void
         }
 
         if (!Utils::isEmpty($catUrls)) {
-            $catUrlAttribute->setValues(array_unique($catUrls));
+            $catUrlAttribute = new Attribute('cat_url');
+            $catUrlAttribute->setValues(Utils::flat($catUrls));
             $this->attributes[] = $catUrlAttribute;
```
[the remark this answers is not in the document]
TheKeymaster: There was no code change regarding this.

tests/Struct/FindologicProductTest.php
```diff
@@ -267,21 +274,22 @@ public function testProductCategoriesUrlWithoutSeoOrEmptyPath(array $data, strin
...
-            new XMLItem('123')
+            new XMLItem('123'),
+            $config
         );
 
         $this->assertTrue($findologicProduct->hasAttributes());
         $attribute = current($findologicProduct->getAttributes());
         $this->assertSame('cat_url', $attribute->getKey());
-        $this->assertContains('/Findologic-Category', $attribute->getValues());
+        $this->assertContains('/FINDOLOGIC-Category/', $attribute->getValues());
```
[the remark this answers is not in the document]
TheKeymaster: Something must have changed. Please check what and how. We must export the same data as we did before this change.

tests/Struct/FindologicProductTest.php
```diff
@@ -1250,24 +1281,24 @@ public function testEmptyCategoryNameShouldStillExportCategory(): void
...
 
         $this->assertCount(6, $findologicProduct->getAttributes());
         $this->assertSame('cat_url', $findologicProduct->getAttributes()[0]->getKey());
 
         $catUrls = $findologicProduct->getAttributes()[0]->getValues();
-        $this->assertCount(2, $catUrls);
-
-        $this->assertSame($expectedFirstCatUrl, $catUrls[0]);
-        $this->assertSame($expectedSecondCatUrl, $catUrls[1]);
+        $this->assertCount(1, $catUrls);
```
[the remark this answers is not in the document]
TheKeymaster: How was this before and how did it change?

tests/Struct/FindologicProductTest.php
```diff
@@ -1932,4 +1966,182 @@ public function testSalesFrequencyIsBasedOnPreviousMonthsOrder(
...
+            $this->getContainer(),
+            $this->shopkey,
+            [],
+            new XMLItem('123'),
+            $config
+        );
+
+        $this->assertTrue($findologicProduct->hasAttributes());
+        $attributes = $findologicProduct->getAttributes();
+        if ($integrationType === 'Direct Integration') {
```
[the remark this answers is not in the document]
TheKeymaster: This comment was only resolved partly. Please also ensure to instead of having the `else`, do a separate check for the `$expectedCategories`.

src/Utils/Utils.php
TheKeymaster: But do we still need it? Can't we get the data in a proper format already, so we don't need this? I imagine this could be very time consuming in case there are a lot of categories.

zaifastafa: We need the seo url of a product recursively along with its parents. So in the recursion we get an array of arrays. Which is then flattened out in the end outside the loop. So this is not called inside the recursion loop.

src/Struct/FindologicProduct.php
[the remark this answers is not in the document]
zaifastafa: There was. An additional check is added in the condition.

src/Struct/FindologicProduct.php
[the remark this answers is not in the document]
zaifastafa: Fetching services from container is not ideal. It is always recommended to use DI instead of container. We would have to mock the whole container just to pass our Config mock. Besides, even if we set the mocked config in container, it does not work, because the sales channel already initializes the config service so it cannot be replaced. 
If we reset it in tearDown it gives the following error in the next test case which needs to use the container. `The previous test case's transaction was not closed properly.` 

the functionality does not really gets affected with this additional argument because we are setting it default as null which in turn always gets from the already initialized container in a real scenario. But in the test cases it makes things much simpler rather than mocking the whole container and all the services it calls. I just do not see a good reason to use the container for getting services, and shopware itself does not recommend that.

tests/Struct/FindologicProductTest.php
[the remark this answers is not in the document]
zaifastafa: OK, previously we were exporting all seo urls, but now we are exporting only the ones associated with the sales channel ID of the shopkey. That is why the count in another test is also different than previous implementation.

tests/Struct/FindologicProductTest.php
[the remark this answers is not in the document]
zaifastafa: See comment above.

src/Struct/FindologicProduct.php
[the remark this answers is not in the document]
TheKeymaster: Fine, I took this point for our future major 3.x release.

TheKeymaster (approved):

PR #199, opened 2021-08-02
PH-297_sw_6_4_3_compatibility (SW-636)

TheKeymaster:


src/Export/Logger/ExportExceptionLogger.php
```diff
@@ -0,0 +1,136 @@
...
+use FINDOLOGIC\FinSearch\Exceptions\Export\Product\ProductHasNoAttributesException;
+use FINDOLOGIC\FinSearch\Exceptions\Export\Product\ProductHasNoCategoriesException;
+use FINDOLOGIC\FinSearch\Exceptions\Export\Product\ProductHasNoNameException;
+use FINDOLOGIC\FinSearch\Exceptions\Export\Product\ProductHasNoPricesException;
+use FINDOLOGIC\FinSearch\Exceptions\Export\Product\ProductInvalidException;
+use Psr\Log\LoggerInterface;
+use Shopware\Core\Content\Product\ProductEntity;
+use Throwable;
+
+class ExportExceptionLogger
```
TheKeymaster, line 19: This little refactoring was done to have easier exception handling and logging. It wouldn't deserve its own story as the change is a bit minor but makes life easier for debugging, as the exception itself is now part of the logged thing as well. This was only done before for `InvalidProductException`, but now is done for all exceptions that may be thrown.

dblaichinger (changes requested):

dblaichinger (approved):

PR #200, opened 2021-08-03
SW-634_fix_duplicate_range_slider

TheKeymaster:


src/Resources/app/storefront/src/js/filter-slider-range.plugin.js
```diff
@@ -27,6 +27,8 @@ export default class FilterSliderRange extends FilterBasePlugin {
   });
 
   init() {
+    this.resetState();
```
dblaichinger, line 30: Wouldn't it be better to check if we have an instance of `noUiSlider` and not skip the whole init in case one exists?

dblaichinger (changes requested):

src/Resources/app/storefront/src/js/filter-slider-range.plugin.js
TheKeymaster: > ... and not skip the whole init in case one exists
We are not skipping the initialization, we're cleaning up any previously rendered content of the container where the last instance of the noUiSlider was, as we are always appending the new noUiSlider to an element which contains it.

Creating a new one on each re-rendering is also the better choice as it always ensures that we have the correct state and the input fields are also binded to the instance of that slider.

dblaichinger: > as we are always appending the new noUiSlider to an element which contains it.

Correct, this is exactly what can be skipped when one already exists.

> ensures that we have the correct state and the input fields are also binded to the instance of that slider.

But the existing instance also ensures correct state and binding, right?

dblaichinger: Discussed in Slack, it's the Shopware-way to initialize everything again. Comment is resolved.

dblaichinger (approved):

PR #217, opened 2021-11-02
SW-612 The exported price must be based on the configured currency

D3xter7:


zaifastafa (changes requested):

tests/Export/ExportPriceTest.php
```diff
@@ -0,0 +1,141 @@
...
+            $context
+        );
+
+        return $currencyId;
+    }
+
+    protected function setup():void{
+        $this->crossSellCategories = ["2221211212121121212122121121212"];
+        $this->logger = new Logger('fl_test_logger');
+    }
```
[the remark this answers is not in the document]
D3xter7: I am using "getExport" function inside my currency class and it is not being inherited from any external class.
`protected function getExport(): XmlExport
    {
        /** @var Router $router */
        $router = $this->getContainer()->get(Router::class);

        return new XmlExport(
            $router,
            $this->getContainer(),
            $this->logger,
            $this->crossSellCategories
        );
    }`

zaifastafa (changes requested):

tests/Export/ExportPriceTest.php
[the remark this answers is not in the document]
zaifastafa: Please don't resolve points that are related to discussions.

tests/Export/ExportPriceTest.php
```diff
@@ -0,0 +1,131 @@
...
+    use ProductHelper;
+    use SalesChannelHelper;
+
+    /** @var FinSearchConfigEntity */
+    private $finsearchConfigEntity;
+
+    /**
+     * @method ContainerInterface getContainer();
+     */
+
```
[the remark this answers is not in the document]
D3xter7: I have some functions like "createNewCurrency" that are not being inherited from any class but directly declare inside "this" class, as this is not being inherited from anywhere so it won't get a container..

src/Export/SalesChannelService.php
```diff
@@ -58,10 +61,18 @@ public function getSalesChannelContext(
         /** @var FinSearchConfigEntity $systemConfigEntity */
         foreach ($systemConfigEntities as $systemConfigEntity) {
             if ($systemConfigEntity->getConfigurationValue() === $shopkey) {
+                $options = [
+                    SalesChannelContextService::LANGUAGE_ID => $systemConfigEntity->getLanguageId(),
+                ];
+                if (method_exists($currentContext, 'getCurrencyId')) {
```
TheKeymaster, line 67: Use `$currentContext->getSalesChannel()->getCurrencyId()`. This way you do not need this `method_exists` check.

src/Struct/FindologicProduct.php
```diff
@@ -795,26 +795,40 @@ protected function setVariantPrices(): void
...
-                    $price->setValue($item->getNet(), $userGroupHash);
-                }
+        foreach ($this->customerGroups as $customerGroup) {
+            $userGroupHash = Utils::calculateUserGroupHash($this->shopkey, $customerGroup->getId());
+            if (Utils::isEmpty($userGroupHash)) {
+                continue;
+            }
 
-                $prices[] = $price;
+            if (!$currencyPrice) {
```
TheKeymaster, line 812: This check should be done before the `foreach` loop.

tests/Traits/DataHelpers/SalesChannelHelper.php
```diff
@@ -61,19 +59,26 @@ public function buildSalesChannelContext(
...
+    ): array {
         $options = [];
         if ($customerEntity) {
             $options[SalesChannelContextService::CUSTOMER_ID] = $customerEntity->getId();
         }
         if ($languageId) {
             $options[SalesChannelContextService::LANGUAGE_ID] = $languageId;
         }
+        // Currency is not available in the sales channel context below Shopware v6.4.
+        if ($currencyId && !Utils::versionLowerThan('6.4')) {
```
TheKeymaster, line 79: The currency id already exists since the first Shopware 6 version. https://github.com/shopware/platform/blob/v6.0.0+dp1/src/Core/System/SalesChannel/Context/SalesChannelContextService.php. The difference is that the helper is only available since a later version, so you need to get the SalesChannel and assign the currency id to the sales channel.

tests/Export/XmlExportTest.php
```diff
@@ -98,4 +100,33 @@ protected function getExport(): XmlExport
...
+        $items = $this->getExport()->buildItems(
+            [$testProduct],
+            self::VALID_SHOPKEY,
+            $customerGroupEntities
+        );
+
+        $item = $items[0];
+        $price = $item->getPrice();
+        $priceValues = $price->getValues();
+        $this->assertEquals(7.5, current($priceValues));
```
TheKeymaster, line 130: How can this pass for versions below 6.4, if you don't set a currency id?

TheKeymaster (changes requested):

tests/Export/XmlExportTest.php
zaifastafa: Because we do not necessarily need the currency ID from the context, as you pointed out in the previous comment. In the code we take the currency ID from the sales channel, which is why the tests pass because we do not need the context currency ID. Have fixed that.

zaifastafa (approved):

TheKeymaster (approved):

PR #218, opened 2021-11-03
SW-654_add_topseller_only_once

TheKeymaster:


tests/Findologic/Api/SortingServiceTest.php
```diff
@@ -70,12 +74,74 @@ public function testProperLegacySortingIsApplied(Request $request, string $expec
...
+        $event = new ProductListingCriteriaEvent(new Request(), new Criteria(), $this->buildSalesChannelContext());
+        $searchRequestHandlerMock = $this->getMockBuilder(NavigationRequestHandler::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $event->getCriteria()->addExtension('sortings', $sortings);
+        $sortingService->handleRequest($event, $searchRequestHandlerMock);
+
+        $actualSortings = $event->getCriteria()->getExtension('sortings')->getElements();
+        $this->assertCount(count($expectedSortings), $actualSortings);
```
dblaichinger, line 129: Can this be just asserted by count and not by name? E.g. assert(SortingService::TOPSELLER_SORT_FIELD, $actualSorting);

dblaichinger (changes requested):

src/Findologic/Api/SortingService.php
```diff
@@ -75,6 +80,23 @@ protected function addTopResultSorting(ProductListingCriteriaEvent $event): void
...
+                    continue;
+                }
+
+                return true;
+            }
+
+            return false;
+        });
+
+        return $topsellerSortings !== [];
```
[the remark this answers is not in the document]
TheKeymaster: No because we use `array_filter`, so the array will contain all topseller sortings. It would return true if there is at least one topseller sorting.

tests/Findologic/Api/SortingServiceTest.php
TheKeymaster: Unfortunately we can not, as internally in the plugin we still make use of `_score` sorting, so this would differ. I also didn't wanted to overcomplicate it. We should be fine ensuring that always one sorting is returned if there is none or if there already is a topseller sorting.

dblaichinger (approved):

PR #222, opened 2021-11-17
SW-660_consider_main_variant_id

TheKeymaster:


src/Export/ProductService.php
```diff
@@ -261,4 +244,47 @@ protected function addProductIdFilters(Criteria $criteria, string $productId): v
...
+
+            $products->add($product);
+        }
+
+        return $products;
+    }
+
+    protected function getRealMainProductWithVariants(string $realMainProductId): ProductEntity
+    {
+        $newProduct = $this->getVisibleProducts(1, 0, $realMainProductId)->first();
```
dblaichinger, line 283: $newProduct could be renamed to $mainProduct

tests/Export/ProductServiceTest.php
```diff
@@ -237,4 +238,97 @@ public function testVariantsWithSameParentButDifferentDisplayGroupAreExportedAsS
...
+            'name' => 'FINDOLOGIC VARIANT 2',
+            'options' => [
+                ['id' => $secondOptionId]
+            ],
+        ]));
+
+        $this->getContainer()->get('product.repository')->update([
+            [
+                'id' => $expectedFirstVariantId,
+                'mainVariantId' => $expectedMainVariantId
```
dblaichinger, line 311: It's not possible to set the mainVariationId directly at createVisibleTestProduct?

src/Export/ProductService.php
```diff
@@ -261,4 +244,47 @@ protected function addProductIdFilters(Criteria $criteria, string $productId): v
...
+
+        $children = $this->getChildrenOrSiblings($newProduct);
+        $newProduct->setChildren($children);
```
dblaichinger, line 286: Those two lines are also used in buildProductsWithVariantInformation and could be removed here to avoid duplicate code.

tests/Export/ProductServiceTest.php
```diff
@@ -237,4 +238,97 @@ public function testVariantsWithSameParentButDifferentDisplayGroupAreExportedAsS
...
+                            'name' => 'Color',
+                        ],
+                    ],
+                ],
+            ],
+            'configuratorGroupConfig' => [
+                [
+                    'id' => $optionGroupId,
+                    // Explicitly set this to false. This tells Shopware to consider the mainVariationId (if set).
+                    'expressionForListings' => false,
```
dblaichinger, line 282: For my understanding, what does expressionForListings mean?

dblaichinger (changes requested):

tests/Export/ProductServiceTest.php
TheKeymaster: No because at this point the related product does not exist yet, which causes an SQL constraint violation.

tests/Export/ProductServiceTest.php
TheKeymaster: It depends on if the variant should be split or not.
image

src/Export/ProductService.php
```diff
@@ -261,4 +244,50 @@ protected function addProductIdFilters(Criteria $criteria, string $productId): v
...
+    {
+        $products = new ProductCollection();
+
+        /** @var ProductEntity $product */
+        foreach ($result->getEntities() as $product) {
+            if ($product->getMainVariantId() !== null) {
+                $product = $this->getRealMainProductWithVariants($product->getMainVariantId());
+            } else {
+                $this->assignChildrenOrSiblings($product);
+            }
```
dblaichinger, line 272: if ($product->getMainVariantId() !== null) {
                $product = $this->getRealMainProductWithVariants($product->getMainVariantId());
            }
            $this->assignChildrenOrSiblings($product);

src/Export/ProductService.php
```diff
@@ -261,4 +244,50 @@ protected function addProductIdFilters(Criteria $criteria, string $productId): v
...
+            $products->add($product);
+        }
+
+        return $products;
+    }
+
+    protected function getRealMainProductWithVariants(string $realMainProductId): ProductEntity
+    {
+        $mainProduct = $this->getVisibleProducts(1, 0, $realMainProductId)->first();
+        $this->assignChildrenOrSiblings($mainProduct);
```
dblaichinger, line 283: Still possible to remove when the if above is changed like suggested.

dblaichinger (changes requested):

src/Export/ProductService.php
TheKeymaster: Ah this is why you wanted it to be changed. Honestly I have rewritten this whole code about three times to get a good structure, so when I got to this point I didn't see that we could simplify this even further. Nice, thanky.

dblaichinger (approved):

dblaichinger (approved):

PR #229, opened 2021-11-30
SW-633 Export search keywords instead of tags

D3xter7:


tests/Struct/FindologicProductTest.php
```diff
@@ -35,6 +35,7 @@
 use Shopware\Core\Content\Category\CategoryCollection;
 use Shopware\Core\Content\Media\Aggregate\MediaThumbnail\MediaThumbnailCollection;
 use Shopware\Core\Content\Media\Aggregate\MediaThumbnail\MediaThumbnailEntity;
+use Shopware\Core\Content\Product\Aggregate\ProductSearchKeyword\ProductSearchKeywordCollection;
```
zaifastafa, line 38: Unused import.

tests/Struct/FindologicProductTest.php
```diff
@@ -363,9 +364,15 @@ public function testProductPrices(?Price $price, ?string $exception): void
 
     public function testProduct(): void
     {
-        $productEntity = $this->createTestProduct();
+        $productId = '29d554327a16fd51350688cfa9930b29';
+        $languageId = Defaults::LANGUAGE_SYSTEM;
+        $productEntity = $this->createTestProduct([
+            'searchKeywords' => [
+                ['id' => Uuid::randomHex(), 'productId' => $productId, 'languageId' => $languageId, 'keyword' => 'FINDOLOGIC Keyword', 'ranking' => 500]
```
zaifastafa, line 371: Run `composer lint` to ensure there are no errors before pushing. For e.g. in this line it seems there are more than allowed characters in a single line. So please fix that.

tests/Struct/FindologicProductTest.php
```diff
@@ -363,9 +364,15 @@ public function testProductPrices(?Price $price, ?string $exception): void
 
     public function testProduct(): void
     {
-        $productEntity = $this->createTestProduct();
+        $productId = '29d554327a16fd51350688cfa9930b29';
```
zaifastafa, line 367: Where is this ID coming from? How can you ensure that this ID is the same product ID that is created for the following product?

src/Struct/FindologicProduct.php
```diff
@@ -356,11 +356,11 @@ public function getKeywords(): array
...
     {
-        $tags = $this->product->getTags();
-        if ($tags !== null && $tags->count() > 0) {
-            foreach ($tags as $tag) {
-                if (!Utils::isEmpty($tag->getName())) {
-                    $this->keywords[] = new Keyword($tag->getName());
+        $keywords = $this->product->getSearchKeywords();
+        if ($keywords !== null && $keywords->count() > 0) {
+            foreach ($keywords as $keyword) {
+                if (!Utils::isEmpty($keyword)) {
```
zaifastafa, line 362: Here the `$keyword` is an object and will never be empty as it is being used inside the loop. We need to ensure the actual keyword text inside the object is not empty, so it needs to check for the keyword value.

zaifastafa (changes requested):
@D3xter7 Make sure firsdt the build passes. After that look at the review changes.

tests/Struct/FindologicProductTest.php
```diff
@@ -390,13 +389,20 @@ public function testProduct(): void
             $config
         );
 
+        $productKeyword = new Keyword('FINDOLOGIC001');
+        $keywords = $findologicProduct->getKeywords();
+        $this->assertNotCount(0, $keywords);
+        foreach ($keywords as $keyword) {
+            if ($keyword->getValue() === $productKeyword->getValue()) {
+                $this->assertEquals($productKeyword, $keyword);
```
zaifastafa, line 397: Having this inside the condition does not sure ensure it gets called. The test would still pass even if this condition is not met. Put a `boolean` flag as `false` by default outside the `foreach`  and if the condition is met, make it `true`.  Then after the `foreach` ensure that the flag is `assertTrue()`. That way we can be sure that the keyword was found.

zaifastafa (changes requested):

tests/Struct/FindologicProductTest.php
```diff
@@ -390,13 +389,23 @@ public function testProduct(): void
...
+        }
+
         $urlBuilderService = $this->getContainer()->get(UrlBuilderService::class);
         $urlBuilderService->setSalesChannelContext($this->salesChannelContext);
 
         $expectedUrl = $urlBuilderService->buildProductUrl($productEntity);
         $this->assertEquals($expectedUrl, $findologicProduct->getUrl());
         $this->assertEquals($productEntity->getName(), $findologicProduct->getName());
-        $this->assertEquals([$productTag], $findologicProduct->getKeywords());
+        $this->assertEquals(true, $productKeyMatched);
```
zaifastafa, line 408: use `assertTrue` here.

tests/Struct/FindologicProductTest.php
```diff
@@ -390,13 +389,23 @@ public function testProduct(): void
             $config
         );
 
+        $productKeyword = new Keyword('FINDOLOGIC001');
+        $productKeyMatched = false;
```
zaifastafa, line 393: a more suitable name would be `expectedProductKeywordExists`

tests/Struct/FindologicProductTest.php
```diff
@@ -390,13 +389,23 @@ public function testProduct(): void
             $config
         );
 
+        $productKeyword = new Keyword('FINDOLOGIC001');
```
zaifastafa, line 392: Where are you assigning the keyword to the product? We have already established that the product number and vendor name is returned in the keywords by default, so we cannot test it against the product number, because the test is not checking if our configured keyword is getting returned or not. So instead of checking against default values, assign a random keyword to the product and make sure that is returned. 
Theoretically this is fine too since we are getting the keyword back, but I guess @TheKeymaster can have a better opinion on that.

zaifastafa (changes requested):

tests/Traits/DataHelpers/ProductHelper.php
```diff
@@ -39,6 +39,7 @@ public function createTestProduct(
         $newCategoryId = Uuid::randomHex();
         $redId = Uuid::randomHex();
         $colorId = Uuid::randomHex();
+        $languageId = Defaults::LANGUAGE_SYSTEM;
```
zaifastafa, line 42: If it is not used anymore please remove it.

tests/Struct/FindologicProductTest.php
```diff
@@ -378,6 +378,14 @@ public function testProduct(): void
         $ordernumbers = $this->getOrdernumber($productEntity);
         $properties = $this->getProperties($productEntity);
 
+        $productSearchKeywordEntity = new ProductSearchKeywordEntity();
+        $productSearchKeywordCollection = new ProductSearchKeywordCollection();
+
+        $productSearchKeywordEntity->setKeyword('FINDOLOGIC Keyword');
+        $productSearchKeywordCollection->set('searchKeyword',$productSearchKeywordEntity);
```
zaifastafa, line 385:
```suggestion
        $productSearchKeywordEntity = new ProductSearchKeywordEntity();
        $productSearchKeywordEntity->setKeyword('FINDOLOGIC Keyword');
        $productSearchKeywordCollection = new ProductSearchKeywordCollection([$productSearchKeywordEntity]);
```

src/Struct/FindologicProduct.php
```diff
@@ -356,11 +356,11 @@ public function getKeywords(): array
 
     protected function setKeywords(): void
     {
-        $tags = $this->product->getTags();
-        if ($tags !== null && $tags->count() > 0) {
-            foreach ($tags as $tag) {
-                if (!Utils::isEmpty($tag->getName())) {
-                    $this->keywords[] = new Keyword($tag->getName());
+        $keywords = $this->product->getSearchKeywords();
```
zaifastafa, line 359: @D3xter7 We need to remove the vendor name and product number from here as well.

zaifastafa (changes requested):

tests/Struct/FindologicProductTest.php
```diff
@@ -390,13 +396,33 @@ public function testProduct(): void
...
+        $ignoreProductKeywordExists = false;
+        $this->assertNotCount(0, $keywords);
+        foreach ($keywords as $keyword) {
+            if ($productKeyword->getValue() === $keyword->getValue()) {
+                $expectedProductKeywordExists = true;
+            }
+            if (in_array($keyword->getValue(), $ignoreKeyords)) {
+                $ignoreProductKeywordExists = true;
+            }
+        }
```
zaifastafa, line 416: As we are now manually ignoring the unwanted keywords here, I think we can directly test for the keywords array instead of checking for just one value inside of it. This will help us in future if some other keyword is also added by Shopware then it will fail here.

This way we can ensure only our configured keywords are being sent in export. Otherwise if another keyword is added we would not know as this test would still pass. 

Also add multiple keywords for that case so there is more than one value to check.

tests/Struct/FindologicProductTest.php
```diff
@@ -390,13 +396,33 @@ public function testProduct(): void
             $config
         );
 
+        $productKeyword = new Keyword('FINDOLOGIC Keyword');
+        $ignoreKeyords = [
+            $productEntity->get('id'),
```
zaifastafa, line 401: Are we getting the product ID as well in the keywords? I thought it was just manufacturer name and product number?

src/Struct/FindologicProduct.php
```diff
@@ -356,11 +356,17 @@ public function getKeywords(): array
 
     protected function setKeywords(): void
     {
-        $tags = $this->product->getTags();
-        if ($tags !== null && $tags->count() > 0) {
-            foreach ($tags as $tag) {
-                if (!Utils::isEmpty($tag->getName())) {
-                    $this->keywords[] = new Keyword($tag->getName());
+        $ignoreKeywords = [$this->product->getProductNumber(), $this->product->get('id'), $this->product
+        ->getManufacturer()->getName()];
```
zaifastafa, line 360: - Use `$manufacturer->getTranslation('name')` instead of getting name directly, see 
`\FINDOLOGIC\FinSearch\Struct\FindologicProduct::setVendors` method for reference. 
- No need to use `$product->get('id')`, you can use `getId()` directly, provided that we need that in the ignored list in the first place. See other comment about it.

src/Struct/FindologicProduct.php
```diff
@@ -356,11 +356,17 @@ public function getKeywords(): array
...
+        $keywords = $this->product->getSearchKeywords();
+
+        if ($keywords !== null && $keywords->count() > 0) {
+            foreach ($keywords as $keyword) {
+                if (!Utils::isEmpty($keyword->getKeyword())) {
+                    $ignoreKeyword = in_array($keyword->getKeyword(), $ignoreKeywords);
```
zaifastafa, line 366: Instead of doing `$keyword->getKeyword()` in both lines, it is efficient to use variable instead.

tests/Struct/FindologicProductTest.php
```diff
@@ -378,6 +378,12 @@ public function testProduct(): void
         $ordernumbers = $this->getOrdernumber($productEntity);
         $properties = $this->getProperties($productEntity);
 
+        $productSearchKeywordEntity = new ProductSearchKeywordEntity();
+        $productSearchKeywordEntity->setId('2255566389092766485905746366');
+        $productSearchKeywordCollection = new ProductSearchKeywordCollection([$productSearchKeywordEntity]);
+        $productSearchKeywordEntity->setKeyword('FINDOLOGIC Keyword');
+        $productEntity->setSearchKeywords($productSearchKeywordCollection);
+
```
zaifastafa, line 386: Extract this into a separate private method so it returns a collection.

zaifastafa (changes requested):

zaifastafa (approved):

tests/Traits/DataHelpers/ProductHelper.php
```diff
@@ -39,7 +39,6 @@ public function createTestProduct(
         $newCategoryId = Uuid::randomHex();
         $redId = Uuid::randomHex();
         $colorId = Uuid::randomHex();
-
```
TheKeymaster, line 42: Please re-add this newline. Why was this even removed?

TheKeymaster (changes requested):

src/Struct/FindologicProduct.php
```diff
@@ -356,11 +356,18 @@ public function getKeywords(): array
 
     protected function setKeywords(): void
     {
-        $tags = $this->product->getTags();
-        if ($tags !== null && $tags->count() > 0) {
-            foreach ($tags as $tag) {
-                if (!Utils::isEmpty($tag->getName())) {
-                    $this->keywords[] = new Keyword($tag->getName());
+        $expelledKeywords = [$this->product->getProductNumber(), $this->product
```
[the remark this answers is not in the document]
TheKeymaster: This comment was only partially resolved.

src/Struct/FindologicProduct.php
```diff
@@ -356,11 +356,18 @@ public function getKeywords(): array
...
+        ->getManufacturer()->getTranslation('name')];
+        $keywords = $this->product->getSearchKeywords();
+
+        if ($keywords !== null && $keywords->count() > 0) {
+            foreach ($keywords as $keyword) {
+                $keywordValue = $keyword->getKeyword();
+                if (!Utils::isEmpty($keywordValue)) {
+                    $expelledKeywordExists = in_array($keywordValue, $expelledKeywords);
```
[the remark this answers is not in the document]
TheKeymaster: This was not resolved.

src/Utils/Utils.php
```diff
@@ -77,6 +77,7 @@ public static function addProductAssociations(Criteria $criteria): Criteria
                 'categories.seoUrls',
                 'translations',
                 'tags',
+                'searchKeywords',
```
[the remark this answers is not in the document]
TheKeymaster: This was not resolved.

tests/Struct/FindologicProductTest.php
```diff
@@ -361,11 +363,18 @@ public function testProductPrices(?Price $price, ?string $exception): void
         }
     }
 
+    private function getKeywordEntity($keyword)
```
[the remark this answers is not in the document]
TheKeymaster: Please remove the PHPdocblock and add the typings directly to the method itself.

src/Struct/FindologicProduct.php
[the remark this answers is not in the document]
TheKeymaster: Please do this everywhere, also in the tests.

TheKeymaster (approved):

PR #232, opened 2021-12-06
SW-507_add_storefront_unit_tests

TheKeymaster:


src/Resources/app/storefront/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "main": "index.js",
+  "directories": {
+    "test": "test"
+  },
+  "scripts": {
+    "test": "jest"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/findologic/plugin-shopware-6.git"
```
dblaichinger, line 14:
```suggestion
    "url": "https://github.com/findologic/plugin-shopware-6.git"
```

src/Resources/app/storefront/test/filter-category-select.plugin.test.js
```diff
@@ -0,0 +1,64 @@
...
+    const checkboxSelector = document.createElement('input');
+    checkboxSelector.classList.add('filter-category-select-checkbox');
+
+    const mockElementButton = document.createElement('button');
+    mockElementButton.classList.add('filter-panel-item-toggle');
+
+    mockElement.appendChild(cmsElementProductListingWrapper);
+    mockElement.appendChild(mockElementButton);
+    mockElement.appendChild(mockElementSpan);
+    mockElement.appendChild(checkboxSelector);
```
dblaichinger, line 36: Can `appendChild` be chained?

src/Resources/app/storefront/test/filter-category-select.plugin.test.js
```diff
@@ -0,0 +1,64 @@
...
+    filterCategorySelectPlugin = new FilterCategorySelectPlugin(mockElement);
+  });
+
+  afterEach(() => {
+    filterCategorySelectPlugin = null;
+  });
+
+  test('filter category select plugin exists', () => {
+    expect(typeof filterCategorySelectPlugin).toBe('object');
+  });
```
dblaichinger, line 63: Maybe add a simple test checking the elements and class in mockElement?

dblaichinger (changes requested):

src/Resources/app/storefront/test/filter-category-select.plugin.test.js
TheKeymaster: It can be chained, but the return value is the appended child and not the `mockElement`.

https://developer.mozilla.org/en-US/docs/Web/API/Node/appendChild

image

src/Resources/app/storefront/test/filter-category-select.plugin.test.js
TheKeymaster: This is the simplest test. Nothing happens to the elements, they're the same as they've been defined in the `beforeEach` and at this point we would be testing testing logic.

src/Resources/app/storefront/package.json
TheKeymaster: Sure why not, it was autogenerated that way so I'd assume it would still be correct.

dblaichinger (approved):

PR #233, opened 2021-12-08
SW-613-exported-main-variant-config

zaifastafa:


src/Export/ProductService.php
```diff
@@ -26,33 +28,45 @@ class ProductService
...
 
     /** @var SalesChannelContext|null */
     private $salesChannelContext;
 
-    public function __construct(ContainerInterface $container, ?SalesChannelContext $salesChannelContext = null)
-    {
+    public function __construct(
+        ContainerInterface $container,
+        ?SalesChannelContext $salesChannelContext = null,
+        ?Config $config = null
```
TheKeymaster, line 43: This is really important, as otherwise this wouldn't be backwards compatible :+1:

src/Export/ProductService.php
```diff
@@ -26,33 +28,45 @@ class ProductService
...
-            $productService = new ProductService($container, $salesChannelContext);
+            $productService = new ProductService($container, $salesChannelContext, $config);
             $container->set(self::CONTAINER_ID, $productService);
         }
 
         if ($salesChannelContext && !$productService->getSalesChannelContext()) {
             $productService->setSalesChannelContext($salesChannelContext);
         }
 
+        if ($config && !$productService->getConfig()) {
```
TheKeymaster, line 66: When would this be the case, shouldn't the constructor already set this? Would we even need a `setConfig` and `getConfig` method?

src/Findologic/MainVariant.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace FINDOLOGIC\FinSearch\Findologic;
+
+final class MainVariant
```
TheKeymaster, line 7: Feels like these constants should belong in the `\FINDOLOGIC\FinSearch\Struct\Config` class.

src/Struct/Config.php
```diff
@@ -63,6 +65,9 @@ class Config extends Struct
     /** @var string */
     private $filterPosition;
 
+    /** @var string */
+    private $mainVariant;
```
TheKeymaster, line 69: Also add this to the constant `ALLOW_FOR_SERIALIZATION`.

src/Struct/Config.php
```diff
@@ -233,4 +244,20 @@ public function getFilterPosition(): string
     {
         return $this->filterPosition;
     }
+
+    /**
+     * @return string
```
TheKeymaster, line 249: Unnecessary docblock.

src/Struct/Config.php
```diff
@@ -233,4 +244,20 @@ public function getFilterPosition(): string
...
+     */
+    public function getMainVariant(): string
+    {
+        return $this->mainVariant;
+    }
+
+    /**
+     * @param string $mainVariant
```
TheKeymaster, line 257: Unnecessary docblock.

tests/Export/ProductServiceTest.php
```diff
@@ -421,4 +423,224 @@ public function testProductIsNotFoundIfMainVariantIsNotAvailableForSalesChannel(
...
+                'parentPrice' => 15,
+                'cheapest' => 2
+            ],
+            'export main parent' => [
+                'config' => 'parent',
+                'parentId' => $parentId,
+                'parentPrice' => 15,
+                'cheapest' => 2
+            ],
+            'export cheapest variant' => [
```
TheKeymaster, line 444: How can this export the cheapest variant, when the `config` is set to `parent`? Also the whole data provider does not seem to have a single case for `cheapest` configuration. How are these cases covered?

tests/Export/ProductServiceTest.php
```diff
@@ -421,4 +423,224 @@ public function testProductIsNotFoundIfMainVariantIsNotAvailableForSalesChannel(
...
+                'config' => 'default',
+                'parentId' => $parentId,
+                'parentPrice' => 15,
+                'cheapest' => 2
+            ],
+            'export main parent' => [
+                'config' => 'parent',
+                'parentId' => $parentId,
+                'parentPrice' => 15,
+                'cheapest' => 2
```
TheKeymaster, line 442: Please always make sure that the data provider names match the method arguments, otherwise this gets a bit confisuing. `cheapest` would be `cheapestPrice`.

tests/Export/ProductServiceTest.php
```diff
@@ -421,4 +423,224 @@ public function testProductIsNotFoundIfMainVariantIsNotAvailableForSalesChannel(
...
+        }
+
+        $mockedConfig = $this->getFindologicConfig(['mainVariant' => $config]);
+        $mockedConfig->initializeBySalesChannel($this->salesChannelContext);
+
+        $this->defaultProductService->setConfig($mockedConfig);
+        $result = $this->defaultProductService->searchVisibleProducts(20, 0);
+        $elements = $result->getElements();
+
+        $this->assertNotEmpty($elements);
```
TheKeymaster, line 609: The moment you assert that the count is at least `1`, you're already ensuring that this is not empty, so I assume this check isn't really necessary. This check can also probably be removed from `testProductWithoutVariantsBasedOnExportConfig`.

tests/Traits/DataHelpers/ConfigHelper.php
```diff
@@ -102,10 +103,32 @@ private function getDefaultFindologicConfigServiceMock(
...
+                        $languageId,
+                        $config['mainVariant']
                     ]
                 ]
             );
 
         return $configServiceMock;
     }
+
+    public function getFindologicConfig(array $override = [], bool $isDirectIntegration = true): Config
```
TheKeymaster, line 119: Why does `$isDirectIntegration` default to `true`? Most of the time we want to ensure that the API behavior is the correct one, so it would make sense to default this to `false`.

tests/Export/ProductServiceTest.php
```diff
@@ -421,4 +423,224 @@ public function testProductIsNotFoundIfMainVariantIsNotAvailableForSalesChannel(
...
+            'name' => 'FINDOLOGIC VARIANT 3',
+            'price' => [['currencyId' => Defaults::CURRENCY, 'gross' => 4, 'net' => 2, 'linked' => false]],
+            'options' => [
+                ['id' => $thirdOptionId]
+            ],
+        ]));
+
+        // Explicitly assign the main variant ID to the product, and assert that the assigned main variant
+        // is the exported variant.
+        if ($config === 'parent') {
```
TheKeymaster, line 579: It is quite hard to read and understand everything this test is checking. Please try to move the test setup into helper methods and things that happen in the `if` blocks into separate tests. In this it probably also makes sense to split the data provider into three separate ones. It is always better to have many small tests instead of one big one. 

In an ideal world a test should also have only one or two assertions (but this is more like a rule of thumb).

src/Export/ProductService.php
```diff
@@ -26,33 +28,45 @@ class ProductService
 {
     public const CONTAINER_ID = 'fin_search.product_service';
 
+    /** @var Config|null */
+    private $config;
```
TheKeymaster, line 32: How could this ever be `null`? If it would be, the container should take care of the resolution (`$container->get(Config::class)`).

TheKeymaster (changes requested):

tests/Export/ProductServiceTest.php
zaifastafa: Ah yes, forgot to change the config. Will fix it.

src/Findologic/MainVariant.php
zaifastafa: We have done a similar thing for the filter position, so was just following the same pattern.

src/Struct/Config.php
zaifastafa: Already exists.

src/Findologic/MainVariant.php
TheKeymaster: Fine :+1:

tests/Traits/DataHelpers/ConfigHelper.php
zaifastafa: The default config we check for everywhere is based on Direct Integration being true, as even in the default config mock which we create (`\FINDOLOGIC\FinSearch\Tests\Traits\DataHelpers\ConfigHelper::getDefaultFindologicConfigServiceMock`), the default integration type is set as Direct Integration. So this was also done here in accordance with that.

src/Export/ProductService.php
zaifastafa: yes it does set this in a real live scenario, but this is following the sales channel pattern. It especially is helpful in tests where we do not have to create the whole instance of the service again and can just work with a mocked config. This if block is most probably unnecessary yes, since this would never be null eventually, but the `get/set` methods would be needed still.

src/Export/ProductService.php
```diff
@@ -275,7 +295,37 @@ protected function buildProductsWithVariantInformation(EntitySearchResult $resul
...
+
+        foreach ($products as $product) {
+            switch ($mainVariantConfig) {
+                case MainVariant::MAIN_PARENT:
+                    $parent = $this->getParentByMainProduct($product);
+                    break;
+                case MainVariant::CHEAPEST:
+                    $parent = $this->getParentByCheapestVariant($product);
+                    break;
+                default:
```
TheKeymaster, line 318: This should throw an exception, as this is an unknown config option. Throwing an `InvalidArgumentException` should be sufficient here.

src/Export/ProductService.php
```diff
@@ -288,4 +338,60 @@ protected function assignChildrenOrSiblings(ProductEntity $product): void
...
+
+        $product->setChildren($children);
+
+        return $product;
+    }
+
+    private function getParentByCheapestVariant(ProductEntity $product): ProductEntity
+    {
+        $currencyId = $this->salesChannelContext->getSalesChannel()->getCurrencyId();
+        // Add the current product as a child, so it gets considered in the loop.
```
TheKeymaster, line 359: Be rather explicit then implicit. Define what ".. the loop" is.

src/Export/ProductService.php
```diff
@@ -288,4 +338,60 @@ protected function assignChildrenOrSiblings(ProductEntity $product): void
         $children = $this->getChildrenOrSiblings($product);
         $product->setChildren($children);
     }
+
+    private function getParentByMainProduct(ProductEntity $product): ProductEntity
```
TheKeymaster, line 342: Change all `private` methods in this class to be `protected`, otherwise extensions can not call them.

TheKeymaster (changes requested):

TheKeymaster (approved):

zaifastafa:
build will pass once #239 is merged

TheKeymaster (approved):

PR #237, opened 2021-12-23
SW-644 Exported categories and cat_urls must also include categories of all variants

D3xter7:


src/Struct/FindologicProduct.php
```diff
@@ -748,14 +748,24 @@ protected function setOrdernumberByProduct(ProductEntity $product): void
     protected function setCategoriesAndCatUrls(): void
     {
         $productCategories = $this->product->getCategories();
-        if ($productCategories === null || empty($productCategories->count())) {
+        $productVariants = $this->product->getChildren()->getElements();
+
+        if (($productCategories === null || empty($productCategories->count()))) {
```
D3xter7, line 753: This condition is inside of the bracket because I had one more condition here to add.

tests/Struct/FindologicProductTest.php
```diff
@@ -186,6 +187,95 @@ public function testNoProductCategories(): void
...
+            if (!empty(array_search($category, $categoriesToBeChecked))) {
+                $isUnique = false;
+                break;
+            }
+            $categoriesArray[] = $category;
+        }
+        $this->assertTrue($isUnique);
+    }
+
+    public function testSetVariantCategoryOnly()
```
zaifastafa, line 234: @D3xter7 What is this test doing? There is nothing being asserted here. Please make some assertion that tell us what categories are returned.

tests/Struct/FindologicProductTest.php
```diff
@@ -186,6 +187,95 @@ public function testNoProductCategories(): void
...
+            $productEntity,
+            $this->router,
+            $this->getContainer(),
+            $this->shopkey,
+            [],
+            new XMLItem('123')
+        );
+    }
+
+    public function testSetParentCategoryOnly()
```
zaifastafa, line 265: @D3xter7 Same here, nothing being asserted here.

tests/Struct/FindologicProductTest.php
```diff
@@ -186,6 +187,95 @@ public function testNoProductCategories(): void
...
+        } else {
+            $catValues = $attributes[0]->getValues();
+        }
+        foreach ($catValues as $category) {
+            if (!empty(array_search($category, $categoriesToBeChecked))) {
+                $isUnique = false;
+                break;
+            }
+            $categoriesArray[] = $category;
+        }
```
zaifastafa, line 230: @D3xter7 This does not guarantee that the values you are checking are of the categories. This can be any other attribute as well. Please do some valid assertions here too.

zaifastafa (changes requested):

tests/Struct/FindologicProductTest.php
D3xter7: I only needed to check if findologic product is being exported.

zaifastafa: @D3xter7 Click-test is something different than Unit Tests. In click-test you need to ensure that the product is exported, but here you are writing unit tests, so it should test the change which you introduced, which in this case is the categories. Please check the test case section on Youtrack it is clearly mentioned to ensure the configured categories are exported properly.

src/Struct/FindologicProduct.php
```diff
@@ -762,14 +762,30 @@ protected function setOrdernumberByProduct(ProductEntity $product): void
     protected function setCategoriesAndCatUrls(): void
     {
         $productCategories = $this->product->getCategories();
-        if ($productCategories === null || empty($productCategories->count())) {
+        $productVariants = $this->product->getChildren();
```
TheKeymaster, line 765: The naming `$productCategories` makes sense here, as this only includes the categories assigned directly to the product. However only a product can have any variants, therefore the naming `$productVariants` is too much information. Calling it `$variants` would be sufficient.

src/Struct/FindologicProduct.php
```diff
@@ -762,14 +762,30 @@ protected function setOrdernumberByProduct(ProductEntity $product): void
...
+
+        // If there are no product categories, and no children either, we throw ProductHasNoCategoriesException.
```
TheKeymaster, line 767: This comment is rather useless, reading the line below is already clear enough. If it wasn't, or if you aren't sure someone else would get what it does, maybe refactor this into a small method with a clearer name. For me personally the code is already self-explanatory.

src/Struct/FindologicProduct.php
```diff
@@ -762,14 +762,30 @@ protected function setOrdernumberByProduct(ProductEntity $product): void
...
+        if ($productCategories->count() === 0 && $productVariants->count() === 0) {
```
TheKeymaster, line 768: You should count the categories of the variants, not the variants themselves. Right now this could cause an issue when a product has no categories, but at least one variant. It would pass, but even tho neither the parent nor any of the variants has a category assigned. Shouldn't there be already a test for this? If not, please create one, otherwise please let me know the name, so I can check what went wrong here.

src/Struct/FindologicProduct.php
```diff
@@ -762,14 +762,30 @@ protected function setOrdernumberByProduct(ProductEntity $product): void
...
             throw new ProductHasNoCategoriesException($this->product);
         }
 
         $catUrls = [];
         $categories = [];
 
         $this->parseCategoryAttributes($productCategories->getElements(), $catUrls, $categories);
+
+        if ($productVariants) {
```
TheKeymaster, line 777: This thuthiness will always pass, as an empty `ProductCollection` instance is never considered falsy. Only if it was `null`, this wouldn't pass, however we set the proper association in the `ProductService` already, therefore this will never be `null`. Consider `$productVariants->count() > 0`.

src/Struct/FindologicProduct.php
```diff
@@ -762,14 +762,30 @@ protected function setOrdernumberByProduct(ProductEntity $product): void
...
+            foreach ($productVariants as $variant) {
+                $variantCategories = $variant->getCategories();
+                if (!$variantCategories) {
```
TheKeymaster, line 780: This is similar to the check in line 777. Consider using the `count()` method of the collection.

src/Utils/Utils.php
```diff
@@ -435,6 +441,6 @@ public static function flat(array $array): array
             $flattened[] = $a;
         });
 
-        return $flattened;
+        return array_unique($flattened, SORT_REGULAR);
```
TheKeymaster, line 444: A unique check violates the description of the docblock. It shouldn't be the responsibility of this method to sort out any non-unique values. The user of this method should be responsible for this.

src/Struct/FindologicProduct.php
```diff
@@ -984,8 +1000,11 @@ protected function parseCategoryAttributes(
                 $this->navigationCategory
             );
 
+
             if (!Utils::isEmpty($categoryPath)) {
-                $categories = array_merge($categories, [$categoryPath]);
+                if (!in_array($categoryPath, $categories)) {
```
TheKeymaster, line 1005: Here it already seems to be too late for this. Can this check be done earlier?

tests/Struct/FindologicProductTest.php
```diff
@@ -189,6 +188,86 @@ public function testNoProductCategories(): void
...
+                'isParentAssigned' => true,
+                'isVariantAssigned' => false
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider parentAndChildrenCategoryProvider
+     */
+    public function testOnlyUniqueCategoriesAreExported(bool $isParentAssigned, bool $isVariantAssigned): void
```
TheKeymaster, line 212: Where are the test cases for

* Create a product with 5 variants. Only assign a category to one of the variants and ensure it is exported properly.
* Remove the set category of the previously created variant. Calling the export now should result in the product being skipped. Ensure that the logs contain information that a ProductHasNoCategoriesException is thrown.

tests/Struct/FindologicProductTest.php
```diff
@@ -189,6 +188,86 @@ public function testNoProductCategories(): void
...
+            ],
+            'Parent has no categories and children have some categories assigned' => [
+                'isParentAssigned' => false,
+                'isVariantAssigned' => true
+            ],
+            'Parent has categories and children have no categories assigned' => [
+                'isParentAssigned' => true,
+                'isVariantAssigned' => false
+            ],
+        ];
```
TheKeymaster, line 206: A data provider should hold `data`, as the name suggests. Here you are simply holding `boolean` values to change the behavior of the test. This is not the purpose of a data provider. Change this data provider to return the product data instead of boolean values, or refactor this into multiple test cases, whatever is easier and makes more sense in the context.

TheKeymaster (changes requested):

src/Struct/FindologicProduct.php
zaifastafa: No because we are only get the category path right before this check. Any where earlier than here, we might not have the correct path that we are building.

tests/Struct/FindologicProductTest.php
zaifastafa: Already replied on Youtrack for this. This is a click test scenario. The unit tests cover the cases.

tests/Struct/FindologicProductTest.php
zaifastafa: The data is the same that we need to pass, so it does not make sense to clutter the provider with the same array of data. We have used providers like this in several other places too.

src/Struct/FindologicProduct.php
```diff
@@ -1042,4 +1059,17 @@ protected function decodeHtmlEntity($value)
 
         return html_entity_decode($value);
     }
+
+    /**
+     * Checks if the product, or any of its children has any category assigned.
+     */
+    private function hasCategories(): bool
```
TheKeymaster, line 1066: Please make this method `protected`, so it can be called by extensions.

src/Struct/FindologicProduct.php
```diff
@@ -1042,4 +1059,17 @@ protected function decodeHtmlEntity($value)
...
+    {
+        $productCategories = $this->product->getCategories();
+        $children = $this->product->getChildren()->filter(function (ProductEntity $variant) {
```
TheKeymaster, line 1069: These are actually `$childrenWithCategories`, but very nice `filter` usage :+1:

src/Utils/Utils.php
```diff
@@ -448,4 +454,9 @@ public static function flat(array $array): array
 
         return $flattened;
     }
+
+    public static function flattenWithUnique(array $array): array
+    {
+        return array_unique(self::flat($array), SORT_REGULAR);
```
TheKeymaster, line 460: Always try to use `static` when in non-private context. When uploading the plugin to the Shopware Store, Shopware's linter doesn't like this, and throws a warning at us :smile:

tests/Struct/FindologicProductTest.php
TheKeymaster: Yeah it's just bad, because we can't extend this test in the future, without making changes in the test itself. Maybe we can refactor this in the future, it was just a minor point anyway.

TheKeymaster (approved):

zaifastafa (approved):

PR #239, opened 2022-01-07
SW-671 Composer 2 runtime api fix

zaifastafa:


src/Utils/Utils.php
```diff
@@ -149,15 +150,23 @@ public static function getCleanShopwareVersion(): string
...
     {
-        $packageVersions = Versions::VERSIONS;
-        $coreIsInstalled = isset($packageVersions['shopware/core']);
-
-        $shopwareVersion = $coreIsInstalled ?
-            Versions::getVersion('shopware/core') :
-            Versions::getVersion('shopware/platform');
-
-        if (!trim($shopwareVersion, 'v@')) {
-            return $coreIsInstalled ? $packageVersions['shopware/core'] : $packageVersions['shopware/platform'];
```
TheKeymaster, line 160: In the past there had been issues with using `Versions::getVersion('shopware/platform')`, therefore getting them directly from the `$packageVersions` is the safer option. See [this commit](https://github.com/findologic/plugin-shopware-6/commit/c463da773960c8125b8d06a6451e09032f127a18) also, in which this has been explicitly opted against.

src/Utils/Utils.php
```diff
@@ -149,15 +150,23 @@ public static function getCleanShopwareVersion(): string
...
+        // Composer 2 runtime API uses the `InstalledVersions::class` in favor of the
+        // deprecated/removed `Versions::class`
+        if (class_exists(InstalledVersions::class)) {
+            $platformIsInstalled = InstalledVersions::isInstalled('shopware/platform');
+            if ($platformIsInstalled) {
+                $shopwareVersion = InstalledVersions::getPrettyVersion('shopware/platform');
+            } else {
+                $shopwareVersion = InstalledVersions::getPrettyVersion('shopware/core');
+            }
```
TheKeymaster, line 161: You could also return the versions here directly, which would result in the `else` block in line 162 to no longer be necessary.

TheKeymaster (changes requested):

src/Resources/public/administration/js/fin-search.js
```diff
@@ -1 +1 @@
...
...          :shopkeyErrorState="shopkeyErrorState"\n                                            :selectedSalesChannelId="selectedSalesChannelId">\n                                    </findologic-config>\n                                {% endblock %}\n                            {% endblock %}\n\n                            {% block findologic_content_card_loading %}\n                                <sw-loader v-if="isLoading"></sw-loader>\n                            {% endblock %}\n                        {% endblock %}\n                    </sw-card-view>\n                {% endblock %}\n            </template>\n        {% endblock %}\n    </sw-page>\n{% endblock %}\n'},gv2C:function(e,t){e.exports="{% block sw_extension_card_base_context_menu_settings %}\n    <sw-context-menu-item\n            v-if=\"isInstalled && extension.configurable\"\n            :routerLink=\"{ name: extension.name === 'FinSearch' ? 'findologic.module.index' : 'sw.extension.config', params: { namespace: extension.name } }\">\n\n        {{ $tc('sw-extension-store.component.sw-extension-card-base.contextMenu.config') }}\n    </sw-context-menu-item>\n{% endblock %}\n"}},[["HerF","runtime","vendors-node"]]]);
```
TheKeymaster, line 1: How did this change? Please revert this change.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #243, opened 2022-01-26
SW-677_encode_html_entities

TheKeymaster:


tests/Struct/FindologicProductTest.php
```diff
@@ -2232,4 +2232,34 @@ public function testCategoryAndCatUrlExportBasedOnIntegrationType(
             $this->assertSame($expectedCategories, $attributes[0]->getValues());
         }
     }
+
+    public function testAttributesAreHtmlEntityEncoded(): void
+    {
+        $expectedAttributeValue = '>80';
+        $productEntity = $this->createTestProduct([
+            'customFields' => [
+                'findologic_yeet' => '&gt;80',
```
dblaichinger, line 2241:
```suggestion
                'length' => '&gt;80',
```

src/Struct/FindologicProduct.php
```diff
@@ -1012,4 +1020,26 @@ protected function getAttributeKey(?string $key): ?string
...
+     * @param mixed $value
+     * @return string|mixed
+     */
+    protected function decodeHtmlEntity($value)
+    {
+        if (!is_string($value)) {
+            return $value;
+        }
+
+        return html_entity_decode($value);
```
dblaichinger, line 1043: The fix is fine, but shouldn't `libflexport` be handling such cases already?

dblaichinger (changes requested):

src/Struct/FindologicProduct.php
TheKeymaster: Why would it? It can be a legitimate case, for an own, or different shopsystem and it limits the developer. Maybe as an option, but even that feels out-of-scope for a library that wraps data into a Findologic-consumable XML/CSV.

tests/Struct/FindologicProductTest.php
TheKeymaster: I liked `findologic_yeet` :cry:

src/Struct/FindologicProduct.php
dblaichinger: Now we need to make sure to call method `decodeHtmlEntity` for every Attribute creation. If a new attribute gets added we might miss it, if libflexport handles it, there is no way to forget it, right?

tests/Struct/FindologicProductTest.php
dblaichinger: Just for readability, otherwise I don't care ;)

src/Struct/FindologicProduct.php
TheKeymaster: Yes but it is in our responsibility to create a wrapper around libflexport to ensure this is always being done. I really don't see this job at libflexport.

dblaichinger (approved):

PR #246, opened 2022-01-31
SW-678_fix_console_errors_for_di

TheKeymaster:


.github/workflows/phpunit.yml
```diff
@@ -58,10 +58,18 @@ jobs:
           ref: ${{ matrix.shopware-versions }}
           path: development
 
+      - name: Add dependency conflicts
+        run: |
+          cd development
+          mv composer.json temp.json
+          jq -r '.conflict."symfony/framework-bundle" |= ">=5.3.14"' temp.json > composer.json
+          rm temp.json
```
dblaichinger, line 66: Why do we need this temp.json? Wouldn't it be possible to add this directly to composer.json?

dblaichinger (changes requested):

.github/workflows/phpunit.yml
TheKeymaster: I tried this, but this is actual the solution using `jq`. If you don't create a temporary file, the whole contents of the file will be removed. There is an issue for that open since 2013 https://github.com/stedolan/jq/issues/105, and the basic solution is to simply create a temporary file.

dblaichinger (approved):

PR #247, opened 2022-02-04
SW-681_shopware_version_detection

dblaichinger:


src/Utils/Utils.php
```diff
@@ -158,24 +158,31 @@ public static function getCleanShopwareVersion(): string
...
     {
+        $version = null;
         // Composer 2 runtime API uses the `InstalledVersions::class` in favor of the
         // deprecated/removed `Versions::class`
         if (class_exists(InstalledVersions::class)) {
-            $platformIsInstalled = InstalledVersions::isInstalled('shopware/platform');
-            if ($platformIsInstalled) {
-                return InstalledVersions::getPrettyVersion('shopware/platform');
+            if (InstalledVersions::isInstalled('shopware/platform')) {
+                $version = $version ?: InstalledVersions::getPrettyVersion('shopware/platform');
```
TheKeymaster, line 166: Since PHP 7.4 you can use `??=`, however since Shopware 6.2 is based on PHP 7.2/7.3, we can only use `??`:

PHP >= 7.1
```suggestion
                $version = $version ?? InstalledVersions::getPrettyVersion('shopware/platform');
```

PHP >= 7.4
```suggestion
                $version ??= InstalledVersions::getPrettyVersion('shopware/platform');
```

src/Utils/Utils.php
```diff
@@ -158,24 +158,31 @@ public static function getCleanShopwareVersion(): string
...
         }
 
-        $packageVersions = Versions::VERSIONS;
-        $platformIsInstalled = isset($packageVersions['shopware/platform']);
-        if ($platformIsInstalled) {
-            return $packageVersions['shopware/platform'];
+        if (defined('PackageVersions\Versions::VERSIONS')) {
+            $packageVersions = Versions::VERSIONS;
+            if (isset($packageVersions['shopware/platform'])) {
+                $version = $version ?: $packageVersions['shopware/platform'];
```
TheKeymaster, line 177: Overriding an already set variable, is screaming for hard debugging time. Please always prefer an early-return or declare a different variable, but don't change the value of an variable that already has a value set before. 

Here is a great article I have found about reassigning variable values https://zellwk.com/blog/dont-reassign/. It talks about JavaScript, but the same principle applies to PHP and many other languages aswell.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #249, opened 2022-02-14
SW-687_add_shopware_6_4_8_1_compatibility

TheKeymaster:


.github/workflows/phpunit.yml
```diff
@@ -59,6 +60,12 @@ jobs:
           path: development
 
       - name: Add dependency conflicts
+        if: |
+          matrix.shopware-versions == 'v6.4.3.1' ||
+          matrix.shopware-versions == 'v6.4.4.0' ||
+          matrix.shopware-versions == 'v6.4.5.1' ||
+          matrix.shopware-versions == 'v6.4.6.0' ||
+          matrix.shopware-versions == 'v6.4.7.0'
```
dblaichinger, line 68: Interesting, do you know why this is necessary only for specific versions?

TheKeymaster: It has something to do how Shopware works internally with dependencies. Since 6.4.3.0 https://github.com/shopware/development/compare/v6.4.2.1...v6.4.3.0 Shopware is committing the `composer.lock` too. However Symfony released framework-bundle v5.3.14 a few weeks ago https://github.com/symfony/framework-bundle/releases/tag/v5.3.14, that caused Shopware to no longer work as intended.
However they've fixed this in Shopware 6.4.8.0, therefore this step is no longer needed for all newer versions.

dblaichinger: Thanks for the explanation!

PR #253, opened 2022-03-01
SW-659 Add separate route for export debug information

zaifastafa:


src/Controller/ProductDebugController.php
```diff
@@ -0,0 +1,218 @@
...
+    private $export;
+
+    /** @var CacheItemPoolInterface */
+    private $cache;
+
+    /** @var SalesChannelService|null */
+    private $salesChannelService;
+
+    /**
+     * @param SalesChannelContextFactory|AbstractSalesChannelContextFactory $salesChannelContextFactory
```
shyim, line 74:
```suggestion
     * @param AbstractSalesChannelContextFactory $salesChannelContextFactory
```

src/Export/ProductService.php
```diff
@@ -169,7 +174,8 @@ protected function getChildrenOrSiblings(ProductEntity $product): ?ProductCollec
 
         // Remove the given children, as the child product is considered as the product, which is shown
         // in the storefront. As we also want to get the data from the parent, we also manually add it here.
-        $children = $result->first()->getChildren();
+        $mainProduct = clone $result->first();
+        $children = $mainProduct->getChildren();
```
TobiasGraml11, line 187: Not using a cloned instance of the product lead to the children being recursively added. So it was possible to do this:
`$result->first()->getChildren()->last()->getChildren()->last()->getChildren()->last()->getChildren()->last()`

Issue came up, because the `JsonResponse` class complained about recursion during JSON parsing:
image

Before:
- Requested Variant
    - Sibling 1
    - Sibling 2
    - Main Variant
        - Sibling 1 / Child 1
        - Sibling 2 / Child 2
        - Main variant
            - .... (and so on)

Now:
- Requested Variant
    - Sibling 1
    - Sibling 2
    - Main Variant
       - Sibling 1 / Child 1
       - Sibling 2 / Child 2

src/Export/ProductDebugService.php
```diff
@@ -0,0 +1,204 @@
...
+namespace FINDOLOGIC\FinSearch\Export;
+
+use FINDOLOGIC\FinSearch\Export\Errors\ExportErrors;
+use FINDOLOGIC\FinSearch\Utils\Utils;
+use Shopware\Core\Content\Product\ProductEntity;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\EntitySearchResult;
+use Symfony\Component\HttpFoundation\JsonResponse;
+
+class ProductDebugService extends ProductService
```
TheKeymaster: This could be a real PITA, since we're restructuring the `ProductService` in the near future, therefore relying on it here would result in more work.

src/Export/ProductDebugService.php
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    private function exportedMainVariantId(): string
+    {
+        $product = $this->fetchProduct($this->productId, true);
+
+        return $product->getId();
+    }
+
+    private function isExported(): bool
```
TheKeymaster: There are many checks in this class, maybe it makes sense to split this up in multiple smaller logical components.

src/Resources/config/services/services.xml
```diff
@@ -44,6 +44,18 @@
             <argument id="serializer.mapping.cache.symfony" type="service" />
         </service>
 
+        <service id="FINDOLOGIC\FinSearch\Controller\ProductDebugController" public="true">
```
TheKeymaster, line 47: Maybe it is easier to add public getters for all properties in the main exporter and inject the `ExportController` as a service here. This way the debug controller isn't as strictly bound to the `ExportController` as it is right now, as basically all the logic needs to be called by the parent, which may lead to unexpected behavior.

src/Validators/DebugExportConfiguration.php
```diff
@@ -0,0 +1,67 @@
...
+     * @Assert\Regex(
+     *     pattern="/^[A-F0-9]{32}$/",
+     *     message="Invalid key provided."
+     * )
+     */
+    private $shopkey;
+
+    /**
+     * @Assert\NotBlank
+     * @Assert\Uuid(
```
TheKeymaster, line 23: Cool, but is this really validating anything?

src/Export/ProductDebugService.php
```diff
@@ -0,0 +1,204 @@
...
+    /** @var ExportErrors */
+    private $exportErrors;
+
+    /** @var ProductEntity */
+    private $product;
+
+    /** @var string[] */
+    private $reasons = [];
+
+    public function getDebugInformation(string $productId, string $shopkey, ExportErrors $exportErrors): JsonResponse
```
TheKeymaster: Maybe split this up into smaller chunks where the sub-classes have their own responsibility.

* Product fetching
* Error collecting
* Mapping data to a JsonResponse

src/Export/ProductDebugService.php
```diff
@@ -0,0 +1,204 @@
...
+        } else {
+            $this->reasons[] = 'Product could not be found or is not available for search.';
+        }
+
+        return false;
+    }
+
+    private function checkExportCriteria(): void
+    {
+        $criteriaFunctions = [
```
TheKeymaster: In this context, maybe `methods` make more sense than `functions`. I personally think of a function more as a callback or something that is defined inline, methods are usually strictly bound to class instances.

src/Export/ProductDebugService.php
```diff
@@ -0,0 +1,204 @@
...
+
+            if ($productErrors = $this->exportErrors->getProductError($this->productId)) {
+                $errors = array_merge($errors, $productErrors->getErrors());
+            }
+        }
+
+        return $errors;
+    }
+
+    private function buildExportUrl(string $exportedMainProductId): string
```
TheKeymaster: Maybe this whole URL building thing could be a own small class.

TheKeymaster (changes requested):

src/Validators/DebugExportConfiguration.php
TobiasGraml11: Yes!
And with `strict=false`, it allows UUIDs without dashes in between.

src/Export/ProductDebugService.php
TobiasGraml11: Leaving this for now, as I am not using functionalities where I expect a lot of changes during variant restructuring.
Would also not be great, to have a lot of duplicated code in those two services

src/Resources/config/services/services.xml
TobiasGraml11: This would also lead to a lot of duplicated code, or a lot of public functions inside the ExportController.
I think it makes sense to add public getters.

TobiasGraml11:
Closing in favor of #285

PR #254, opened 2022-03-04
SW-690 Fix export of products without manufacturer

TobiasGraml11:


tests/Struct/FindologicProductTest.php
```diff
@@ -459,9 +459,25 @@ private function getKeywordEntity(string $keyword): ProductSearchKeywordEntity
         return $productSearchKeywordEntity;
     }
 
-    public function testProduct(): void
+    public function hasManufacturerProvider(): array
     {
-        $productEntity = $this->createTestProduct();
+        return [
+            'Has manufacturer' => [true],
```
dblaichinger, line 465: To increase readability please restructure like:
```
'Product with manufacturer' => [
    'hasManufacturer => true
],
```

dblaichinger (changes requested):

dblaichinger (approved):

src/Struct/FindologicProduct.php
```diff
@@ -358,10 +358,13 @@ protected function setKeywords(): void
     {
         $blackListedKeywords = [
             $this->product->getProductNumber(),
```
TheKeymaster, line 360: Is the product number also nullable?

tests/Traits/DataHelpers/ProductHelper.php
```diff
@@ -31,7 +31,8 @@ public function createVisibleTestProduct(array $overrides = [], bool $withVarian
     public function createTestProduct(
         array $overrideData = [],
         bool $withVariant = false,
-        bool $overrideRecursively = false
+        bool $overrideRecursively = false,
+        bool $withManufacturer = true
```
TheKeymaster, line 35: Wouldn't it be easier to use the `$overrideData`? I want to avoid we add more parameters to this, as a method call already looks completely wild when using all parameters:

```php
$this->createTestProduct([], false, false, false);
```

Without checking the code or method arguments you would have no idea whats going on. If you want to easily create products without a manufacturer, you could add a method with the name `createTestProductWithoutManufacturer` or similar.

tests/Struct/FindologicProductTest.php
```diff
@@ -459,9 +459,29 @@ private function getKeywordEntity(string $keyword): ProductSearchKeywordEntity
         return $productSearchKeywordEntity;
     }
 
-    public function testProduct(): void
+    public function hasManufacturerProvider(): array
     {
-        $productEntity = $this->createTestProduct();
+        return [
+            'Product with manufacturer' => [
+                'hasManufacturer' => true,
```
TheKeymaster, line 466: This is preference, but I typically avoid boolean values for data providers, as with every boolean you have to check the code to grasp what's going on. In this case it is simple, but once we add more things here, it may get hard to read. In this example I would simply set `manufacturerName` to either `FINDOLOGIC` or `null`. It is fine the way it is now, unless you already want to change that, but please just keep this in mind for the future.

TheKeymaster (changes requested):

src/Struct/FindologicProduct.php
TobiasGraml11: No, it is not:
image

tests/Traits/DataHelpers/ProductHelper.php
TobiasGraml11: Setting the manufacturer `null` inside the data array leads to an error while creating:
image

tests/Struct/FindologicProductTest.php
TobiasGraml11: I think this would make more sense, if you could set the manufacturer name `null` and set it in `$overrideData`.
Renaming it, to be consistent with the parameter of `createTestProduct`.

tests/Traits/DataHelpers/ProductHelper.php
TheKeymaster: We need to find a solution for this in the future. Otherwise we'ill end up with a method that is unreadable and hardly usable. A builder pattern would be the best approach here. For now we can leave it like that.

tests/Struct/FindologicProductTest.php
TheKeymaster: See https://github.com/findologic/plugin-shopware-6/pull/254#discussion_r822446617.

tests/Struct/FindologicProductTest.php
```diff
@@ -459,9 +459,25 @@ private function getKeywordEntity(string $keyword): ProductSearchKeywordEntity
...
+            ],
+            'Product without manufacturer' => [
+                'withManufacturer' => false,
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider hasManufacturerProvider
+     * @param bool $withManufacturer
```
TheKeymaster, line 476: Already typehinted parameters don't need a docblock referring to their type.

TheKeymaster (approved):

PR #257, opened 2022-03-14
SW-693 Do not export information of inactive variants

TobiasGraml11:


dblaichinger (changes requested):

src/Export/ProductService.php
```diff
@@ -205,6 +203,16 @@ protected function buildProductCriteria(?int $limit = null, ?int $offset = null)
         return $criteria;
     }
 
+    protected function addVisibilityFilter(Criteria $criteria): void
```
TheKeymaster, line 206: Nice that you already marked it as `protected` :+1:

tests/Export/ProductServiceTest.php
```diff
@@ -117,29 +117,53 @@ public function testFindsVariantForInactiveProduct(): void
         $this->assertSame('FINDOLOGIC VARIANT', $product->getName());
     }
 
-    public function testIfMoreThanOneVariantExistsItWillUseVariantInformationInsteadOfMainProductInformation(): void
+    public function variantProvider(): array
     {
+        return [
+            'variant information is used instead of product information' => [
+                'mainProductActive' => false,
+                'firstVariantActive' => true,
```
TheKeymaster, line 125: `firstVariantActive` may suit this use case, but it is more flexible when you just provide the variants as data. This way it is easier adding new data to the data provider.

```php
$parentId = Uuid::randomHex();

// ...
[
    'variants' => [
        [
            'active' => true,
            'id' => $parentId
            // ...
        ],
        [
            'active' => false,
            'parentId' => $parentId
            // ...
        ]
    ],
    'expectedChildCount' => 1',
];
```

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #258, opened 2022-03-15
PH-552 - Only show categories of sales channel in config

TobiasGraml11:


TobiasGraml11:
Closed in favor of #259 and #260

PR #259, opened 2022-03-16
SW-694 Dynamic product groups categories must work as cross-selling category

TobiasGraml11:


src/Export/XmlExport.php
```diff
@@ -130,16 +130,48 @@ private function exportSingleItem(
         return $xmlProduct->getXmlItem();
     }
 
-    private function isProductInCrossSellingCategory(ProductEntity $productEntity): bool
+    private function isProductInCrossSellingCategory(ProductEntity $productEntity): ?CategoryEntity
```
dblaichinger, line 133: If method name starts with `is` I would expect a boolean return, maybe `getEntityFromCrossSellingCategories`?

dblaichinger: If method name starts with `is` I would expect a boolean return, maybe `getEntityFromCrossSellingCategories`?

src/Export/XmlExport.php
```diff
@@ -130,16 +130,48 @@ private function exportSingleItem(
...
-            $categoryId = $category ? $category->getId() : null;
+        if (count($this->crossSellingCategories)) {
+            $categories = array_merge(
+                $this->getAssignedCategories($productEntity),
+                $this->getDynamicProductGroupCategories($productEntity)
+            );
+
+            foreach ($categories as $categoryId => $category) {
+                if (in_array($categoryId, $this->crossSellingCategories)) {
+                    return $category;
```
dblaichinger, line 143: This would only return first match, is this fine?

src/Export/XmlExport.php
```diff
@@ -130,16 +130,48 @@ private function exportSingleItem(
...
+    private function getDynamicProductGroupCategories(ProductEntity $productEntity): array
+    {
+        $dynamicProductGroupService = null;
+        if ($this->container->has('fin_search.dynamic_product_group')) {
+            $dynamicProductGroupService = $this->container->get('fin_search.dynamic_product_group');
+        }
 
-            return (in_array($categoryId, $this->crossSellingCategories));
+        if ($dynamicProductGroupService) {
+            return $dynamicProductGroupService->getCategories($productEntity->getId());
```
dblaichinger: This could be moved into the upper if.

tests/Traits/DataHelpers/CategoryHelper.php
```diff
@@ -27,9 +28,17 @@ public function createTestCategory(array $categories)
         /** @var CategoryCollection $result */
         $result = $repository->search($criteria, Context::createDefaultContext());
 
-        /** @var CategoryEntity $first */
-        $first = $result->first();
+        return $result->first();
+    }
+
+    public function createBasicCategory(?array $overrideData = []): CategoryEntity
```
dblaichinger: As discussed, please try to combine into `createTestCategory`

dblaichinger (changes requested):

src/Export/XmlExport.php
TobiasGraml11: I would say so, yes.
If needed, I can change it to listing all categories.

dblaichinger (approved):

tests/Export/ProductIdExportTest.php
```diff
@@ -6,17 +6,14 @@
...
 use FINDOLOGIC\FinSearch\Export\ProductIdExport;
 use FINDOLOGIC\FinSearch\Export\XmlExport;
+use Shopware\Core\Content\Category\CategoryEntity;
+use Shopware\Core\Content\Product\ProductEntity;
 use Shopware\Storefront\Framework\Routing\Router;
 
 class ProductIdExportTest extends XmlExportTest
 {
-    public function testProductsInCrossSellCategoriesAreNotWrappedAndErrorIsLogged(): void
+    public function buildItemsAndAssertError(ProductEntity $product, CategoryEntity $category)
```
TheKeymaster, line 15: This method is never called, therefore nothing tests this logic for the product id export.

TheKeymaster (changes requested):

tests/Export/ProductIdExportTest.php
TobiasGraml11: The test class extends `XmlExportTest`, so this is just an extended version of the helper function defined there.
It is called during the tests of `XmlExportTest`.

image

TheKeymaster (approved):

PR #260, opened 2022-03-16
SW-695 Selection of cross-selling categories must not be limited to 500 categories

TobiasGraml11:


TobiasGraml11:
Closed in favor of #263

PR #264, opened 2022-03-21
SW-664 adding adapters

msilinskas:


src/Export/Adapters/DateAddedAdapter.php
```diff
@@ -0,0 +1,25 @@
...
+use FINDOLOGIC\Export\Data\DateAdded;
+use Shopware\Core\Content\Product\ProductEntity;
+
+class DateAddedAdapter
+{
+    public function adapt(ProductEntity $product): ?DateAdded
+    {
+        $releaseDate = $product->getReleaseDate();
+
+        if (null === $releaseDate) {
```
TheKeymaster: Please don't use yoda conditions :D

src/Export/Adapters/NameAdapter.php
```diff
@@ -14,17 +14,28 @@ class NameAdapter
...
 
         return $value;
     }
 
-    protected function getCleanedName(ProductEntity $product): string
+    protected function getCleanedName(ProductEntity $product): ?string
     {
         $name = $product->getTranslation('name');
+
+        if (!$name) {
```
TheKeymaster: Why was this added? If the product has no name an exception should be thrown.

src/Export/Adapters/OrderNumberAdapter.php
```diff
@@ -0,0 +1,64 @@
...
+        if (null === $orderNumber) {
+           return [];
+        }
+
+        $orderNumbers[] = $orderNumber;
+
+        return $orderNumbers;
+    }
+
+    private function getChildrenProductsOrderNumbers(ProductEntity $product): array
```
TheKeymaster: The adapters shouldn't fetch data from children products. Instead (this will be done in a future story) each child should call the `adapt` method to get the required data. For now please make sure no adapter gets any data from children.

src/Export/Adapters/OrderNumberAdapter.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $orderNumbers = [];
+
+        foreach ($product->getChildren() as $productEntity) {
+            $orderNumber = $this->getOrderNumber($productEntity);
+            if (null === $orderNumber) {
```
TheKeymaster: 7182_yoda: Remove you should these yoda expressions. 

Please just check through your code, there may be other occurrences.

src/Export/Adapters/PropertiesAdapter.php
```diff
@@ -0,0 +1,144 @@
...
+use FINDOLOGIC\FinSearch\Utils\Utils;
+use Shopware\Core\Content\Product\ProductEntity;
+use Shopware\Core\Framework\DataAbstractionLayer\Pricing\Price as ProductPrice;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class PropertiesAdapter
+{
+    /** @var array  */
+    private $properties = [];
```
TheKeymaster: Adapters should not be state aware and this is dangerous. If you would call this adapter multiple times, you will pollute it more and more, which not only only gives you incorrect data, but may eventually also cause a memory-leak. Please don't store any data in a service.

src/Export/Adapters/SalesFrequencyAdapter.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public function adapt(ProductEntity $product): ?SalesFrequency
+    {
+        $orders = $this->orderLineItemRepository->searchIds(
+            $this->getCriteria($product),
+            $this->salesChannelContext->getContext()
+        );
+
+        if (!$orders->getTotal()) {
```
TheKeymaster: This would also fail if it was `0` and `getTotal` always returns an `int` anyway, so what's the point of this check?

src/Export/Adapters/SalesFrequencyAdapter.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public function __construct(
+        EntityRepository $orderLineItemRepository,
+        SalesChannelContext $salesChannelContext
+    ) {
+        $this->orderLineItemRepository = $orderLineItemRepository;
+        $this->salesChannelContext = $salesChannelContext;
+    }
+
+    public function adapt(ProductEntity $product): ?SalesFrequency
```
TheKeymaster, line 34: salesfrequency shouldn't be nullable. Either it hasn't sold, which would be `0`, or it has been sold, which would be the count of products sold.

src/Export/Adapters/SalesFrequencyAdapter.php
```diff
@@ -0,0 +1,66 @@
...
+            return null;
+        }
+
+        $salesFrequency = new SalesFrequency();
+        $salesFrequency->setValue($orders->getTotal());
+
+        return $salesFrequency;
+    }
+
+    private function getCriteria(ProductEntity $product): Criteria
```
TheKeymaster: Maybe `buildCriteria` would suit better here.

TheKeymaster (changes requested):

src/Export/Adapters/AdapterFactory.php
```diff
@@ -18,16 +18,56 @@ class AdapterFactory
     /** @var UrlAdapter */
     protected $urlAdapter;
 
+    /** @var DescriptionAdapter */
+    protected $descriptionAdapter;
+
```
TobiasGraml11: I think it would make sense to set this and the adapters above `private` as well

src/Export/Adapters/AttributeAdapter.php
```diff
@@ -12,6 +12,7 @@
 use FINDOLOGIC\FinSearch\Export\UrlBuilderService;
 use FINDOLOGIC\FinSearch\Struct\Config;
 use FINDOLOGIC\FinSearch\Utils\Utils;
+use Psr\Container\ContainerInterface;
```
TobiasGraml11: This import is unused

src/Export/Adapters/DateAddedAdapter.php
```diff
@@ -0,0 +1,25 @@
...
+
+class DateAddedAdapter
+{
+    public function adapt(ProductEntity $product): ?DateAdded
+    {
+        $releaseDate = $product->getReleaseDate();
+
+        if (!$releaseDate) {
+            return null;
+        }
```
TobiasGraml11: This should also work:
```suggestion
        if (!$releaseDate = $product->getReleaseDate()) {
            return null;
        }
```

src/Export/Adapters/DescriptionAdapter.php
```diff
@@ -0,0 +1,37 @@
...
+        $value = new Description();
+        $description = $this->getDescription($product);
+
+        if (!$description) {
+            return null;
+        }
+
+        $value->setValue($description);
+
+        return $value;
```
TobiasGraml11: I think we should stay consistent here, you are creating the `Description` class just before setting the value in the `DateAddedAdapter`.
```suggestion
        if (!$descriptionValue = $this->getDescription($product)) {
            return null;
        }

        $description = new Description();
        $description->setValue($descriptionValue);

        return $description;
```

src/Export/Adapters/ImagesAdapter.php
```diff
@@ -0,0 +1,29 @@
...
+namespace FINDOLOGIC\FinSearch\Export\Adapters;
+
+use FINDOLOGIC\Export\Data\Image;
+use FINDOLOGIC\FinSearch\Export\ProductImageService;
+use Shopware\Core\Content\Product\ProductEntity;
+
+class ImagesAdapter
+{
+    /** @var ProductImageService $productImageService */
+    private $productImageService;
```
TobiasGraml11: This member variables are set to `protected` in the `AttributeAdapter`. I think it makes sense to make it `protected`, so that the service can be used in plugin extensions.

src/Export/Adapters/KeywordsAdapter.php
```diff
@@ -0,0 +1,65 @@
...
+
+declare(strict_types=1);
+
+namespace FINDOLOGIC\FinSearch\Export\Adapters;
+
+use FINDOLOGIC\Export\Data\Keyword;
+use FINDOLOGIC\FinSearch\Utils\Utils;
+use Shopware\Core\Content\Product\Aggregate\ProductSearchKeyword\ProductSearchKeywordCollection;
+use Shopware\Core\Content\Product\ProductEntity;
+use Shopware\Core\System\Tag\TagCollection;
```
TobiasGraml11: Unused import

src/Export/Adapters/KeywordsAdapter.php
```diff
@@ -0,0 +1,65 @@
...
+        $blackListedKeywords = [
+            $product->getProductNumber(),
+        ];
+
+        if ($manufacturer = $product->getManufacturer()) {
+            $blackListedKeywords[] = $manufacturer->getTranslation('name');
+        }
+
+        return $blackListedKeywords;
+    }
```
TobiasGraml11: Please make those functions `protected`, so they can be used and overwritten in extension plugins.
This also relates to other adapters.

src/Export/Adapters/OrderNumberAdapter.php
```diff
@@ -0,0 +1,48 @@
...
+            return new Ordernumber($product->getEan());
+        }
+
+        if (!Utils::isEmpty($product->getManufacturerNumber())) {
+            return new Ordernumber($product->getManufacturerNumber());
+        }
+
+        return null;
+    }
+}
```
TobiasGraml11: We will need to export all 3 types of order numbers. Currently, only the product number would be exported.

```suggestion
    private function getOrderNumber(ProductEntity $product): ?Ordernumber
    {
        $ordernumbers = [];
    
        if (!Utils::isEmpty($product->getProductNumber())) {
            $ordernumbers[] = new Ordernumber($product->getProductNumber());
        }

        if (!Utils::isEmpty($product->getEan())) {
            $ordernumbers[] = new Ordernumber($product->getEan());
        }

        if (!Utils::isEmpty($product->getManufacturerNumber())) {
            $ordernumbers[] = new Ordernumber($product->getManufacturerNumber());
        }

        return $ordernumbers;
    }
}
```

src/Export/Adapters/OrderNumberAdapter.php
```diff
@@ -0,0 +1,48 @@
...
+
+use FINDOLOGIC\Export\Data\Ordernumber;
+use FINDOLOGIC\FinSearch\Utils\Utils;
+use Shopware\Core\Content\Product\ProductEntity;
+
+class OrderNumberAdapter
+{
+    public function adapt(ProductEntity $product): array
+    {
+        return $this->getParentItemOrderNumber($product);
```
TobiasGraml11: I think this function only needs to include the suggested content of `getOrderNumber` from my comment below (or above?). Splitting this into 2 extra functions seems kinda unnecessary.

src/Export/Adapters/ImagesAdapter.php
TobiasGraml11: Please also consider this for the other adapters

src/Export/Adapters/SalesFrequencyAdapter.php
```diff
@@ -0,0 +1,62 @@
...
+
+    public function __construct(
+        EntityRepository $orderLineItemRepository,
+        SalesChannelContext $salesChannelContext
+    ) {
+        $this->orderLineItemRepository = $orderLineItemRepository;
+        $this->salesChannelContext = $salesChannelContext;
+    }
+
+    public function adapt(ProductEntity $product): SalesFrequency
```
TobiasGraml11: > All `adapt()` methods that only return a single instance such as Name, must also be nullable (this way an extension could optionally simply return null)
```suggestion
    public function adapt(ProductEntity $product): ?SalesFrequency
```

src/Export/Adapters/UserGroupsAdapter.php
```diff
@@ -0,0 +1,31 @@
...
+use FINDOLOGIC\FinSearch\Utils\Utils;
+use Shopware\Core\Checkout\Customer\Aggregate\CustomerGroup\CustomerGroupEntity;
+use Shopware\Core\Content\Product\ProductEntity;
+
+class UserGroupsAdapter
+{
+    /**
+     * @return Usergroup[]
+     */
+    public function adapt(ProductEntity $product, ExportContext $exportContext): array
```
TobiasGraml11: Please include the export context via constructor parameter, like it is done in the `AttributeAdapter`.

src/Export/ExportItemAdapter.php
```diff
@@ -55,12 +55,49 @@ public function adapt(Item $item, ProductEntity $product): Item
...
+
+        foreach ($this->adapterFactory->getKeywordsAdapter()->adapt($product) as $keyword) {
+            $item->addKeyword($keyword);
+        }
+
+        foreach ($this->adapterFactory->getImagesAdapter()->adapt($product) as $image) {
+            $item->addImage($image);
+        }
+
+        $item->setSalesFrequency($this->adapterFactory->getSalesFrequencyAdapter()->adapt($product));
```
TobiasGraml11: We should also move this inside a nullable check, as the return type should be nullable

tests/Export/Adapters/AttributeAdapterTest.php
```diff
@@ -0,0 +1,830 @@
...
+use Shopware\Core\Framework\Adapter\Translation\Translator;
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\DataAbstractionLayer\Exception\InconsistentCriteriaIdsException;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
+use Shopware\Core\Framework\Test\TestCaseBase\IntegrationTestBehaviour;
+use Shopware\Core\Framework\Uuid\Uuid;
+use Shopware\Core\System\SalesChannel\Context\SalesChannelContextFactory;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Shopware\Core\System\SalesChannel\SalesChannelEntity;
```
TobiasGraml11: A lot of unused imports

tests/Export/Adapters/AttributeAdapterTest.php
```diff
@@ -0,0 +1,830 @@
...
+
+declare(strict_types=1);
+
+namespace FINDOLOGIC\FinSearch\Tests\Export;
+
+use FINDOLOGIC\Export\Data\Attribute;
+use FINDOLOGIC\Export\Data\Property;
+use FINDOLOGIC\Export\XML\XMLItem;
+use FINDOLOGIC\FinSearch\Exceptions\Export\Product\AccessEmptyPropertyException;
+use FINDOLOGIC\FinSearch\Exceptions\Export\Product\ProductHasNoCategoriesException;
```
TobiasGraml11, line 9: As this exception is an unused import, it suggests that you forgot about the test for it.
This exception was tested in `testProductAndVariantHaveNoCategories`

tests/Export/Adapters/AttributeAdapterTest.php
```diff
@@ -0,0 +1,830 @@
...
+            $this->getContainer()->get(Translator::class),
+            $this->getContainer()->get('fin_search.sales_channel_context'),
+            $this->getContainer()->get(UrlBuilderService::class),
+            $this->getContainer()->get('fin_search.export_context')
+        );
+    }
+
+    /**
+     * @param Attribute[] $attributes
+     * @param array<string, string> $customFields
```
TobiasGraml11: This type does not fit for multidimensional custom fields
```suggestion
     * @param array<string, string|array> $customFields
```

tests/Export/Adapters/AttributeAdapterTest.php
```diff
@@ -0,0 +1,830 @@
...
+    }
+
+    /**
+     * @dataProvider multiSelectCustomFieldsProvider
+     */
+    public function testProductWithMultiSelectCustomFields(
+        array $customFields,
+        array $expectedCustomFieldAttributes
+    ): void {
+        $data['customFields'] = ['long_value' => str_repeat('und wieder, ', 20000)];
```
TobiasGraml11: `$customFields` is never used, did you mean?:
```suggestion
    public function testProductWithMultiSelectCustomFields(
        array $customFields,
        array $expectedCustomFieldAttributes
    ): void {
        $data['customFields'] = $customFields;
```

tests/Export/Adapters/AttributeAdapterTest.php
```diff
@@ -0,0 +1,830 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\Exception\InconsistentCriteriaIdsException;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
+use Shopware\Core\Framework\Test\TestCaseBase\IntegrationTestBehaviour;
+use Shopware\Core\Framework\Uuid\Uuid;
+use Shopware\Core\System\SalesChannel\Context\SalesChannelContextFactory;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Shopware\Core\System\SalesChannel\SalesChannelEntity;
+
+class AttributeAdapterTest extends TestCase
```
TobiasGraml11, line 35: Some tests that seem to be skipped, or at least I can't find them.:
- testProductAndVariantHaveNoCategories
- testOnlyUniqueCategoriesAreExported

Correct me, if I am wrong.

tests/Export/Adapters/DescriptionAdapterTest.php
```diff
@@ -0,0 +1,72 @@
...
+use FINDOLOGIC\Export\Exceptions\EmptyValueNotAllowedException;
+use FINDOLOGIC\Export\XML\XMLItem;
+use FINDOLOGIC\FinSearch\Export\Adapters\DescriptionAdapter;
+use FINDOLOGIC\FinSearch\Export\FindologicProductFactory;
+use FINDOLOGIC\FinSearch\Tests\Traits\DataHelpers\ProductHelper;
+use FINDOLOGIC\FinSearch\Tests\Traits\DataHelpers\SalesChannelHelper;
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\Test\TestCaseBase\IntegrationTestBehaviour;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
```
TobiasGraml11: Unused imports. Please also check the other adapter test classes.

tests/Export/Adapters/AttributeAdapterTest.php
```diff
@@ -0,0 +1,830 @@
+<?php
+
+declare(strict_types=1);
+
+namespace FINDOLOGIC\FinSearch\Tests\Export;
```
TobiasGraml11: This needs to be changed in all adapter test files

```suggestion
namespace FINDOLOGIC\FinSearch\Tests\Export\Adapters;
```

tests/Export/Adapters/PropertiesAdapterTest.php
```diff
@@ -0,0 +1,143 @@
...
+use FINDOLOGIC\FinSearch\Tests\Traits\DataHelpers\ProductHelper;
+use FINDOLOGIC\FinSearch\Tests\Traits\DataHelpers\PropertiesHelper;
+use FINDOLOGIC\FinSearch\Tests\Traits\DataHelpers\SalesChannelHelper;
+use FINDOLOGIC\FinSearch\Utils\Utils;
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Defaults;
+use Shopware\Core\Framework\Test\TestCaseBase\IntegrationTestBehaviour;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+
+class PropertiesAdapterTest extends TestCase
```
TobiasGraml11, line 20: I think your are missing following test:
- testNonFilterablePropertiesAreExportedAsPropertiesInsteadOfAttributes

tests/Traits/DataHelpers/ProductHelper.php
```diff
@@ -295,4 +298,71 @@ public function getBasicVariantData(array $overrides = []): array
...
+
+    protected function getEncodedUrl(string $url): string
+    {
+        $parsedUrl = parse_url($url);
+        $urlPath = explode('/', $parsedUrl['path']);
+        $encodedPath = array_map('\FINDOLOGIC\FinSearch\Utils\Utils::multiByteRawUrlEncode', $urlPath);
+        $parsedUrl['path'] = implode('/', $encodedPath);
+
+        return Utils::buildUrl($parsedUrl);
+    }
```
TobiasGraml11: How about moving this into an `ImageHelper`?

TobiasGraml11 (changes requested):

src/Export/Adapters/DescriptionAdapter.php
```diff
@@ -0,0 +1,35 @@
...
+            return null;
+        }
+
+        $description = new Description();
+        $description->setValue($descriptionValue);
+
+        return $description;
+    }
+
+    private function getDescription(ProductEntity $product): ?string
```
TobiasGraml11: Please make it `protected`

src/Export/Adapters/PropertiesAdapter.php
```diff
@@ -0,0 +1,217 @@
...
+class PropertiesAdapter
+{
+    /** @var SalesChannelContext $salesChannelContext */
+    private $salesChannelContext;
+
+    /** @var TranslatorInterface $translator */
+    private $translator;
+
+    /** @var Config $config */
+    private $config;
```
TobiasGraml11: Please make them `protected`

src/Export/Adapters/SalesFrequencyAdapter.php
```diff
@@ -0,0 +1,62 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\RangeFilter;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+
+class SalesFrequencyAdapter
+{
+    /** @var EntityRepository $orderLineItemRepository */
+    private $orderLineItemRepository;
+
+    /** @var SalesChannelContext $salesChannelContext */
+    private $salesChannelContext;
```
TobiasGraml11: Please make them `protected`

src/Export/Adapters/UserGroupsAdapter.php
```diff
@@ -0,0 +1,39 @@
...
+use FINDOLOGIC\Export\Data\Usergroup;
+use FINDOLOGIC\FinSearch\Export\ExportContext;
+use FINDOLOGIC\FinSearch\Utils\Utils;
+use Shopware\Core\Checkout\Customer\Aggregate\CustomerGroup\CustomerGroupEntity;
+use Shopware\Core\Content\Product\ProductEntity;
+
+class UserGroupsAdapter
+{
+    /** @var ExportContext $exportContext */
+    private $exportContext;
```
TobiasGraml11: Please make it `protected`

src/Export/Adapters/OrderNumberAdapter.php
```diff
@@ -0,0 +1,36 @@
...
+            $orderNumbers[] = new Ordernumber($product->getEan());
+        }
+
+        if (!Utils::isEmpty($product->getManufacturerNumber())) {
+            $orderNumbers[] = new Ordernumber($product->getManufacturerNumber());
+        }
+
+        return $orderNumbers;
+    }
+}
```
TobiasGraml11:
```suggestion
class OrderNumberAdapter
{
    public function adapt(ProductEntity $product): array
    {
        return $this->getOrderNumbers($product);
    }

    protected function getOrderNumbers(ProductEntity $product): array
    {
        $orderNumbers = [];

        if (!Utils::isEmpty($product->getProductNumber())) {
            $orderNumbers[] = new Ordernumber($product->getProductNumber());
        }

        if (!Utils::isEmpty($product->getEan())) {
            $orderNumbers[] = new Ordernumber($product->getEan());
        }

        if (!Utils::isEmpty($product->getManufacturerNumber())) {
            $orderNumbers[] = new Ordernumber($product->getManufacturerNumber());
        }

        return $orderNumbers;
    }
}
```

tests/Export/Adapters/AttributeAdapterTest.php
```diff
@@ -0,0 +1,899 @@
+<?php
+
+declare(strict_types=1);
+
+namespace FINDOLOGIC\FinSearch\Tests\Adapters\Export;
```
TobiasGraml11: This should always correspond to the path.

This is the case for all adapter tests.
```suggestion
namespace FINDOLOGIC\FinSearch\Tests\Export\Adapters;
```

TobiasGraml11 (changes requested):

TobiasGraml11 (approved):

src/Export/Adapters/NameAdapter.php
```diff
@@ -14,17 +14,18 @@ class NameAdapter
...
+    public function adapt(ProductEntity $product): ?Name
     {
         $value = new Name();
         $value->setValue($this->getCleanedName($product));
 
         return $value;
     }
 
-    protected function getCleanedName(ProductEntity $product): string
+    protected function getCleanedName(ProductEntity $product): ?string
```
dblaichinger: In which case can the return value be `null`? The product name is required for the exported data so it would be better to run into ProductHasNoNameException instead

src/Export/Adapters/UrlAdapter.php
```diff
@@ -25,10 +25,14 @@ public function __construct(
         $this->urlBuilderService = $urlBuilderService;
     }
 
-    public function adapt(ProductEntity $product): Url
+    public function adapt(ProductEntity $product): ?Url
     {
         $rawUrl = $this->urlBuilderService->buildProductUrl($product);
 
+        if (!$rawUrl) {
```
dblaichinger: Is this also a valid case? 
`urlBuilderService->buildProductUrl()` also declares non-nullable string as return.

src/Export/ExportItemAdapter.php
```diff
@@ -55,12 +55,52 @@ public function adapt(Item $item, ProductEntity $product): Item
...
         foreach ($this->adapterFactory->getAttributeAdapter()->adapt($product) as $attribute) {
             $item->addMergedAttribute($attribute);
         }
+
         $item->setAllPrices($this->adapterFactory->getPriceAdapter()->adapt($product));
-        $item->setUrl($this->adapterFactory->getUrlAdapter()->adapt($product));
+
+        if ($url = $this->adapterFactory->getUrlAdapter()->adapt($product)) {
+            $item->setUrl($url);
+        }
```
dblaichinger, line 105: Names and urls have been always set before, why do we need the if checks here?

src/Export/Adapters/NameAdapter.php
msilinskas: Yep, it's not necessary here.

TobiasGraml11: @dblaichinger @msilinskas  This is related to this point in the story:
> All adapt() methods that only return a single instance such as Name, must also be nullable (this way an extension could optionally simply return null)

src/Export/ExportItemAdapter.php
TobiasGraml11: @dblaichinger @msilinskas This is related to this point in the story:

> All adapt() methods that only return a single instance such as Name, must also be nullable (this way an extension could optionally simply return null)

src/Export/Adapters/NameAdapter.php
msilinskas: @TobiasGraml11 I guess that only adapt method should be nullable as it is now?

TobiasGraml11: @msilinskas Yes, "child" functions can be whatever they are, as you don't have to extend them.

dblaichinger: @TobiasGraml11 Makes sense, we should make the adapt methods nullable. Let's discuss if we should add an exception for required values and let the product skip in SW-666 then.

PR #265, opened 2022-03-23
SW-698 Fetch Shopware version from Shopware directly

TobiasGraml11:


src/Resources/config/services/services.xml
```diff
@@ -157,7 +157,9 @@
         />
 
         <service id="FINDOLOGIC\FinSearch\Findologic\Api\FindologicSearchService" autowire="true" />
-        <service id="FINDOLOGIC\FinSearch\Findologic\Api\SortingService" autowire="true" />
+        <service id="FINDOLOGIC\FinSearch\Findologic\Api\SortingService" autowire="true">
+            <argument type="string" key="$shopwareVersion">%kernel.shopware_version%</argument>
```
dblaichinger, line 162: Should there be a leading `$` for the key?

`key="$shopwareVersion"`

src/Resources/config/services/services.xml
```diff
@@ -157,7 +157,9 @@
         />
 
         <service id="FINDOLOGIC\FinSearch\Findologic\Api\FindologicSearchService" autowire="true" />
-        <service id="FINDOLOGIC\FinSearch\Findologic\Api\SortingService" autowire="true" />
+        <service id="FINDOLOGIC\FinSearch\Findologic\Api\SortingService" autowire="true">
```
dblaichinger, line 161: Maybe define the service outside of the `FindologicSearchService`?

dblaichinger (changes requested):

src/Resources/config/services/services.xml
TobiasGraml11: What do you mean? This are defined as completely separate services here.

dblaichinger: We can leave it as it is until the service need to be injected somewhere else. It was just a suggestion.

TobiasGraml11: Can you elaborate on your suggestion? :P
I don't get the point at all

dblaichinger: As example see `FindologicConfigService`, defined on line 23 and injected as argument in several other lines.

src/Resources/config/services/services.xml
TobiasGraml11: Yes:
https://symfony.com/doc/current/service_container.html#manually-wiring-arguments

src/Resources/config/services/services.xml
dblaichinger: As discussed in Slack, I read it wrong, but please add empty lines between those services.

src/Resources/config/services/services.xml
dblaichinger: Discussed in Slack, resolved.

dblaichinger (approved):

PR #266, opened 2022-03-28
SW-699 Fix cross-selling categories with non-existent config value

TobiasGraml11:


src/Resources/app/administration/src/module/findologic-module/components/findologic-config/index.js
```diff
@@ -125,12 +125,18 @@ Component.register('findologic-config', {
...
-      this.actualConfigData['FinSearch.config.crossSellingCategories'].push(item.id)
+      if (this.actualConfigData['FinSearch.config.crossSellingCategories']) {
+        this.actualConfigData['FinSearch.config.crossSellingCategories'].push(item.id);
+      } else {
+        this.actualConfigData['FinSearch.config.crossSellingCategories'] = [item.id];
+      }
     },
 
     onCategoryRemove(item) {
+      const config = this.actualConfigData['FinSearch.config.crossSellingCategories'] || [];
```
dblaichinger: Is this empty array case even valid? I'd expect at least one item when `onCategoryRemove` gets called.

src/Resources/app/administration/src/module/findologic-module/components/findologic-config/index.js
```diff
@@ -125,12 +125,18 @@ Component.register('findologic-config', {
     },
 
     onCategoryAdd(item) {
-      this.actualConfigData['FinSearch.config.crossSellingCategories'].push(item.id)
+      if (this.actualConfigData['FinSearch.config.crossSellingCategories']) {
+        this.actualConfigData['FinSearch.config.crossSellingCategories'].push(item.id);
+      } else {
+        this.actualConfigData['FinSearch.config.crossSellingCategories'] = [item.id];
+      }
```
dblaichinger, line 132: Wouldn't it make sense to add / remove entries from `this.categoryCollection` and then store it to actualConfigData? Please explain the purpose of categoryCollection array.

dblaichinger (changes requested):

src/Resources/app/administration/src/module/findologic-module/components/findologic-config/index.js
TobiasGraml11: I just made this to be sure, but I can remove it

src/Resources/app/administration/src/module/findologic-module/components/findologic-config/index.js
TobiasGraml11: It's still just an implementation like SW does it: https://github.com/shopware/platform/blob/v6.4.9.0/src/Administration/Resources/app/administration/src/module/sw-sales-channel/view/sw-sales-channel-detail-base/index.js

The `categoryCollection` is there, because the SW component needs a collection to save the selected categories. There is no easier way to map the collection information to our needed ID array, than by using the provided events

src/Resources/app/administration/src/module/findologic-module/components/findologic-config/index.js
dblaichinger: If it has no purpose, please remove it.

dblaichinger (changes requested):

dblaichinger (approved):

PR #270, opened 2022-04-04
SW-702 Ignore products with price of 0

TobiasGraml11:


src/Export/ProductService.php
```diff
@@ -180,6 +182,7 @@ protected function getCriteriaWithProductVisibility(?int $limit = null, ?int $of
     {
         $criteria = $this->buildProductCriteria($limit, $offset);
         $this->addVisibilityFilter($criteria);
+        $this->addHasPriceFilter($criteria);
```
dblaichinger:
```suggestion
        $this->addPriceZeroFilter($criteria);
```

src/Export/ProductService.php
```diff
@@ -383,12 +395,14 @@ protected function getParentByCheapestVariant(ProductEntity $product): ProductEn
...
         foreach ($children as $child) {
-            $price = $child->getCurrencyPrice($currencyId);
-            if (!$price) {
+            if (!$price = $child->getCurrencyPrice($currencyId)) {
                 continue;
             }
 
-            if ($price->getGross() < $cheapestPrice->getGross()) {
+            if (
+                $cheapestPrice->getGross() === 0.0 ||
```
dblaichinger, line 405: Shouldn't we `continue` when price=0?

tests/Export/ProductServiceTest.php
```diff
@@ -93,6 +93,17 @@ public function testGetInstancePopulatesSalesChannelContext(): void
         $this->assertSame($this->salesChannelContext, $productService->getSalesChannelContext());
     }
 
+    public function testIgnoresProductsWithoutPrice(): void
```
dblaichinger:
```suggestion
    public function testIgnoresProductsWithPriceZero(): void
```

dblaichinger (changes requested):

src/Export/ProductService.php
TobiasGraml11: The cheapest price is defined before the `foreach`, to have some value to start with. The value is initially set to the parent price.
This check just makes sure, that the price is set with a value > 0, in case the main product has no price. At this place, all variants with a price of 0 are already filtered out, so no check needed.

PR #271, opened 2022-04-05
SW-703_product_listing_filter_container

dblaichinger:
…iner before product listing


src/Resources/views/storefront/page/search/layout/leftFilterLayout.html.twig
```diff
@@ -0,0 +1,35 @@
...
+{% endif %}
+
+<div class="row">
+    <div class="cms-element-sidebar-filter {{ filterColumns }}">
+        <div id="filter-panel-wrapper"
+             class="filter-panel-wrapper"
+             data-offcanvas-filter-content="true">
+            {% sw_include '@Storefront/storefront/component/listing/filter-panel.html.twig' with {
+                listing: page.listing,
+                sidebar: isInSidebar
```
TobiasGraml11: This is always true inside of leftFilterLayout, right? So I guess we can remove this parameter now. (Including the definition in search-pagelet)

src/Resources/views/storefront/page/search/layout/topFilterLayout.html.twig
```diff
@@ -0,0 +1,20 @@
+<div class="cms-element-sidebar-filter">
+    <div id="filter-panel-wrapper"
+         class="filter-panel-wrapper"
+         data-offcanvas-filter-content="true">
+        {% sw_include '@Storefront/storefront/component/listing/filter-panel.html.twig' with {
+            listing: page.listing,
+            sidebar: isInSidebar
```
TobiasGraml11: `sidebar` will always need `false` for the top layout

src/Resources/views/storefront/page/search/search-pagelet.html.twig
```diff
@@ -21,39 +19,9 @@
         {% if page.header.extensions.flConfig.filterPosition is constant('\\FINDOLOGIC\\FinSearch\\Findologic\\FilterPosition::LEFT') %}
             {% set sidebarClasses = 'has--sidebar' %}
```
TobiasGraml11: Where is this variable even used?

TobiasGraml11 (changes requested):

src/Resources/views/storefront/page/search/search-pagelet.html.twig
dblaichinger: Used in Shopware templates, but it's also set there so I don't see any reason to do it here.

src/Resources/views/storefront/page/search/search-pagelet.html.twig
```diff
@@ -2,58 +2,23 @@
 
 {% set listingColumns = 'col-sm-6 col-lg-4 col-xl-3' %}
```
TobiasGraml11: When we add all those other variables inside the enabled check, should we also move this there?

TobiasGraml11: Or why is this even needed? If it is something from Shopware, I guess this should also work without us setting it again

dblaichinger: Seems like it's not necessary, classes are still set when removed
Screen Shot 2022-04-07 at 10 08 47
:

TobiasGraml11 (approved):

PR #272, opened 2022-04-06
SW-707 only_use_node_14_for_old_shopware_versions

TheKeymaster:


.github/workflows/phpunit.yml
```diff
@@ -41,6 +41,16 @@ jobs:
...
+        if: |
+          matrix.shopware-versions == 'v6.4.0.0' ||
+          matrix.shopware-versions == 'v6.4.1.0' ||
+          matrix.shopware-versions == 'v6.4.2.1' ||
+          matrix.shopware-versions == 'v6.4.3.1' ||
+          matrix.shopware-versions == 'v6.4.4.0' ||
+          matrix.shopware-versions == 'v6.4.5.1' ||
+          matrix.shopware-versions == 'v6.4.6.0' ||
+          matrix.shopware-versions == 'v6.4.7.0' ||
+          matrix.shopware-versions == 'v6.4.8.2'
```
dblaichinger, line 53: Interesting, why was this not necessary for older versions e.g. `6.3.x.x`?

dblaichinger, line 53: Interesting, why was this not necessary for older versions e.g. `6.3.x.x`?

TheKeymaster: We don't run JS tests for legacy Shopware versions, which includes everything below 6.4. As we don't run any JS tests, we also don't install any JavaScript dependencies, therefore this change wasn't required for `phpunit-legacy.yml`.

dblaichinger (approved):

PR #274, opened 2022-04-07
SW-709 Show filter button on mobile

TobiasGraml11:


src/Resources/views/storefront/page/search/search-pagelet.html.twig
```diff
@@ -16,7 +16,8 @@
         {% if page.header.extensions.flConfig.filterPosition is constant('\\FINDOLOGIC\\FinSearch\\Findologic\\FilterPosition::LEFT') %}
             {% sw_include '@FinSearch/storefront/page/search/layout/leftFilterLayout.html.twig' %}
         {% else %}
-            {% sw_include '@FinSearch/storefront/page/search/layout/topFilterLayout.html.twig' %}
+            {# The top layout is the Shopware default #}
+            {{ parent() }}
```
dblaichinger, line 20: Since we call parent above as well, we could check for `FilterPosition::TOP` above and remove this parent call.

src/Resources/views/storefront/page/search/layout/leftFilterLayout.html.twig
```diff
@@ -10,17 +10,18 @@ is configured on the left side.
...
-             data-offcanvas-filter-content="true">
-            {% sw_include '@Storefront/storefront/component/listing/filter-panel.html.twig' with {
-                listing: page.listing,
-                sidebar: true
-            } %}
-        </div>
-    </div>
+    {% sw_include '@Storefront/storefront/element/cms-element-sidebar-filter.html.twig' with {
+        listing: page.listing,
+        sidebar: 1
```
dblaichinger: Use `true` like Shopware

src/Resources/views/storefront/page/search/layout/leftFilterLayout.html.twig
```diff
@@ -10,17 +10,18 @@ is configured on the left side.
...
-        <div id="filter-panel-wrapper"
-             class="filter-panel-wrapper"
-             data-offcanvas-filter-content="true">
-            {% sw_include '@Storefront/storefront/component/listing/filter-panel.html.twig' with {
-                listing: page.listing,
-                sidebar: true
-            } %}
-        </div>
-    </div>
+    {% sw_include '@Storefront/storefront/element/cms-element-sidebar-filter.html.twig' with {
```
dblaichinger, line 21: Does this show the mobile filter also for < 6.4.4?

dblaichinger (changes requested):

src/Resources/views/storefront/page/search/layout/leftFilterLayout.html.twig
TobiasGraml11: Yes

src/Resources/views/storefront/page/search/search-pagelet.html.twig
TobiasGraml11: This way, we would not set the `dataUrl` and `filterUrl` variables. So I think this structure makes the most sense.

src/Resources/views/storefront/page/search/layout/leftFilterLayout.html.twig
TobiasGraml11: Seems like SW is not really sure themselves, but changed it to boolean on our side

image

src/Resources/views/storefront/page/search/layout/leftFilterLayout.html.twig
dblaichinger: Awesome!

dblaichinger (approved):

PR #276, opened 2022-04-25
SW-666 Products must be skipped if required data is not set

msilinskas:


src/Export/ExportItemAdapter.php
```diff
@@ -35,26 +44,52 @@ class ExportItemAdapter implements ExportItemAdapterInterface
...
     }
 
-    public function adapt(Item $item, ProductEntity $product): Item
+    public function adapt(Item $item, ProductEntity $product): ?Item
     {
         $this->eventDispatcher->dispatch(new BeforeItemAdaptEvent($product, $item), BeforeItemAdaptEvent::NAME);
 
+        try {
+            $item = $this->adaptProduct($item, $product);
+        } catch (EmptyValueNotAllowedException | ProductInvalidException | Exception | Throwable $exception) {
```
TobiasGraml11: Shouldn't `Throwable` catch all exceptions anyway? I think the other types are unnecessary.

tests/Export/ExportItemAdapterTest.php
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    public function testProductInvalidExceptionIsLogged(): void
+    {
+        $id = Uuid::randomHex();
+
+        $productEntity = $this->createTestProduct([
+            'id' => $id,
+            'categories' => []
+        ]);
```
TobiasGraml11: - The ID is never used, as you always get it with `$productEntity->getId()` anyway
- We can just throw the `ProductHasNoCategoriesException` directly. If the exception is thrown at all, is already included in the corresponding adapter test. (Already done on my side)

```suggestion
        $productEntity = $this->createTestProduct();
```

tests/Export/ExportItemAdapterTest.php
```diff
@@ -0,0 +1,179 @@
...
+        );
+    }
+
+    public function testEmptyValueIsNotAllowedExceptionIsLogged(): void
+    {
+        $id = Uuid::randomHex();
+
+        $productEntity = $this->createTestProduct([
+            'id' => $id
+        ]);
```
TobiasGraml11: The ID is never used, as you always get it with `$productEntity->getId()` anyway
```suggestion
         $productEntity = $this->createTestProduct();
```

tests/Export/ExportItemAdapterTest.php
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    public function testThrowableExceptionIsLogged(): void
+    {
+        $id = Uuid::randomHex();
+        $errorMessage = 'This product failed, because it is faulty.';
+
+        $productEntity = $this->createTestProduct([
+            'id' => $id
+        ]);
```
TobiasGraml11: The ID is never used, as you always get it with `$productEntity->getId()` anyway

```suggestion
        $errorMessage = 'This product failed, because it is faulty.';

        $productEntity = $this->createTestProduct();
```

TobiasGraml11 (changes requested):

TobiasGraml11 (approved):

PR #278, opened 2022-05-03
SW-716 Round prices to 2 decimals

TobiasGraml11:


src/Struct/FindologicProduct.php
```diff
@@ -582,8 +582,8 @@ protected function setProperties(): void
                 /** @var ProductPrice $listPrice */
                 $listPrice = $price->getListPrice();
                 if ($listPrice) {
-                    $this->addProperty('old_price', (string)$listPrice->getGross());
-                    $this->addProperty('old_price_net', (string)$listPrice->getNet());
+                    $this->addProperty('old_price', (string)round($listPrice->getGross(), 2));
+                    $this->addProperty('old_price_net', (string)round($listPrice->getNet(), 2));
```
dblaichinger: Would prefer a space between `(string)round`

src/Struct/FindologicProduct.php
```diff
@@ -849,16 +849,16 @@ protected function getPricesFromProduct(ProductEntity $product): array
...
-                $price->setValue($netPrice, $userGroupHash);
+                $price->setValue(round($netPrice, 2), $userGroupHash);
             }
 
             $prices[] = $price;
         }
 
         $price = new Price();
-        $price->setValue($currencyPrice->getGross());
+        $price->setValue(round($currencyPrice->getGross(), 2));
```
dblaichinger, line 861: Since we need to call it for every price entry, how about adding this in libflexport?

tests/Struct/FindologicProductTest.php
```diff
@@ -450,6 +450,51 @@ public function testProductPrices(?Price $price, ?string $exception): void
...
+        return [
+            'Price that is rounded down' => [
+                'expectedPrice' => 15.39,
+                'actualPrice' => 15.3945
+            ],
+            'Price that is rounded up' => [
+                'expectedPrice' => 15.76,
+                'actualPrice' => 15.7591351
+            ],
+            'Price that is rounded to an integer' => [
```
dblaichinger: 20.00 is still a float number.

dblaichinger: 'Price that is rounded to a whole number'

dblaichinger (changes requested):

src/Struct/FindologicProduct.php
TobiasGraml11: Not sure if libflexport should also be used to manipulate data, as it's a simple tool to provide the XML generation.

tests/Struct/FindologicProductTest.php
TobiasGraml11: Translation-wise, an integer is just a number without decimals, no matter how you write it down.
But I'll change it to make it more clear here

src/Struct/FindologicProduct.php
dblaichinger: That's a valid point, with the downside that we might take care of rounding prices in all of our export components. This goes way further than the fix of the story, resolved.

dblaichinger (approved):

PR #281, opened 2022-05-05
SW-704 export variant specific data

msilinskas:


src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,302 @@
...
+        );
+    }
+
+    protected function getVisibleProducts(?int $limit, ?int $offset, ?string $productId): EntitySearchResult
+    {
+        $criteria = $this->getCriteriaWithProductVisibility($limit, $offset);
+
+        $criteria->addFilter(
+            new EqualsFilter('parentId', null)
+        );
```
TobiasGraml11: I would suggest something like:

```suggestion
        $criteria = $this->getBaseCriteria($limit, $offset);
        $this->adaptCriteriaBasedOnConfiguration($criteria)
```

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,302 @@
...
+        $this->addVisibilityFilter($criteria);
+        $this->addPriceZeroFilter($criteria);
+
+        return $criteria;
+    }
+
+
+    protected function buildProductCriteria(?int $limit = null, ?int $offset = null): Criteria
+    {
+        $criteria = new Criteria();
```
TobiasGraml11:
```suggestion
    protected function getBaseCriteria(?int $limit = null, ?int $offset = null): Criteria
    {
        $criteria = $this->getCriteriaWithProductVisibility();
```

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    protected function getParentCriteriaByMainProduct(Criteria $criteria): Criteria
+    {
+        $criteria->addFilter(
+            new EqualsFilter('parentId', null)
+        );
+
+        return $criteria;
+    }
```
TobiasGraml11:
```suggestion
    protected function adaptCriteriaForMainProductConfiguration(Criteria $criteria): Criteria
    {
        $criteria->addFilter(
            new EqualsFilter('parentId', null)
        );
    }
```

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,302 @@
...
+
+    protected function getParentCriteriaByCheapestVariant(Criteria $criteria): Criteria
+    {
+        return $criteria;
+    }
```
TobiasGraml11: You will probably need to check the price of the main product and the cheapest variant after the database call. Except you can think about a way of requesting it directly from the database:

Something like this:
```php
protected function getCheapestVariant(ProductEntity $product): ProductEntity
{
    $currencyId = $this->salesChannelContext->getSalesChannel()->getCurrencyId();
    $cheapestVariant = $product->getChildren()->first();

    $productPrice = $product->getCurrencyPrice(currencyId);
    $cheapestVariantPrice = $cheapestVariant->getCurrencyPrice(currencyId);

    return $productPrice->getGross() < $cheapestVariantPrice->getGross() ? $product : $cheapestVariant;
}
```

```suggestion
    protected function adaptCriteriaForCheapestVariantConfiguration(Criteria $criteria): void
    {
        $children = $criteria->getAssociation('children');
        $children->setLimit(1);
        $children->addSorting(
            new FieldSorting('price', FieldSorting::ASCENDING)
        );
        $this->handleAvailableStock($children);
        $this->addVisibilityFilter($children);
        $this->addPriceZeroFilter($children);
    }
```

Maybe you can even filter out variants, that are more expensive than the main product. This way, you could simplify the check I mentioned above

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,302 @@
...
+                break;
+            case MainVariant::CHEAPEST:
+                $criteria = $this->getParentCriteriaByCheapestVariant($criteria);
+                break;
+            default:
+                throw new InvalidArgumentException($mainVariantConfig);
+        }
+
+       return $criteria;
+    }
```
TobiasGraml11:
```suggestion
    protected function adaptCriteriaBasedOnConfiguration(Criteria $criteria): Criteria
    {
        $mainVariantConfig = $this->config->getMainVariant();

        switch ($mainVariantConfig) {
            case MainVariant::SHOPWARE_DEFAULT:
                break;
            case MainVariant::MAIN_PARENT:
                $this->adaptCriteriaForMainProductConfiguration($criteria);
                break;
            case MainVariant::CHEAPEST:
                $this->adaptCriteriaForCheapestVariantConfiguration($criteria);
                break;
            default:
                throw new InvalidArgumentException($mainVariantConfig);
        }
    }
```

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,302 @@
...
+    {
+        return $this->salesChannelContext;
+    }
+
+    public function getConfig(): ?Config
+    {
+        return $this->config;
+    }
+
+    public function searchVisibleProducts(
```
TobiasGraml11, line 100: This function should return exactly what the function did in `ProductService` (without requesting all the variants). Just the logic behind it should be made much better and clearer.
And what the function did, was returning a list of products according to the plugin configuration:
- default: return the products how SW does it in their code
- main/parent product: Get the main product or the main variant (Can be configured in the product configuration)
- cheapest: Get the cheapest variant. Use the main product, if there is no cheapest

src/Controller/ExportController.php
```diff
@@ -112,11 +113,16 @@ protected function initialize(Request $request, ?SalesChannelContext $context):
...
+        $this->productService = ProductServiceSeparateVariants::getInstance(
             $this->container,
             $this->salesChannelContext,
             $this->pluginConfig
         );
+//        $this->productService = ProductService::getInstance(
+//            $this->container,
+//            $this->salesChannelContext,
+//            $this->pluginConfig
+//        );
```
TobiasGraml11: This can be removed now

src/Controller/ExportController.php
```diff
@@ -151,6 +157,7 @@ protected function doExport(): Response
             $this->exportConfig->getStart(),
             $this->exportConfig->getProductId()
         );
+//        dd($products);
```
TobiasGraml11: This can be removed now

src/Controller/ExportController.php
```diff
@@ -112,11 +113,16 @@ protected function initialize(Request $request, ?SalesChannelContext $context):
 
         $this->container->set('fin_search.sales_channel_context', $this->salesChannelContext);
         $this->pluginConfig = $this->getPluginConfig();
-        $this->productService = ProductService::getInstance(
+        $this->productService = ProductServiceSeparateVariants::getInstance(
```
TobiasGraml11, line 118: Please also update it for the member variable docblock

src/Controller/ExportController.php
```diff
@@ -10,6 +10,7 @@
 use FINDOLOGIC\FinSearch\Export\HeaderHandler;
 use FINDOLOGIC\FinSearch\Export\ProductIdExport;
 use FINDOLOGIC\FinSearch\Export\ProductService;
```
TobiasGraml11: Should not be needed anymore

src/Controller/ExportController.php
```diff
@@ -10,6 +10,7 @@
 use FINDOLOGIC\FinSearch\Export\HeaderHandler;
```
TobiasGraml11, line 10: It seems like the `$salesChannelContextFactory` member variable is not needed at all for this controller.
Could you please remove that? As this will be a major release, I think we can do that.

src/Export/ExportItemAdapter.php
```diff
@@ -38,13 +40,13 @@ class ExportItemAdapter implements ExportItemAdapterInterface
...
-    /** @var AdapterFactory */
+    /** @var AdapterFactory  $adapterFactory*/
     protected $adapterFactory;
 
-    /** @var ExportContext */
+    /** @var ExportContext $exportContext*/
     protected $exportContext;
 
-    /** @var LoggerInterface  */
+    /** @var LoggerInterface  $logger*/
```
TobiasGraml11, line 49: For consistency, please also add it to the other member variables

src/Export/ExportItemAdapter.php
```diff
@@ -83,6 +85,29 @@ public function adapt(Item $item, ProductEntity $product): ?Item
...
+            foreach ($this->adapterFactory->getAttributeAdapter()->adapt($product) as $attribute) {
+                $item->addMergedAttribute($attribute);
+            }
+
+            foreach ($this->adapterFactory->getPropertiesAdapter()->adapt($product) as $property) {
+                $item->addProperty($property);
+            }
+        } catch (Throwable $exception) {
+            return;
+        }
```
TobiasGraml11: I think it makes sense to also log exceptions for variants

src/Export/ExportItemAdapter.php
```diff
@@ -83,6 +85,29 @@ public function adapt(Item $item, ProductEntity $product): ?Item
         return $item;
     }
 
+    public function adaptVariant(Item $item, ProductEntity $product): void
```
TobiasGraml11: We should also return the `Item` here

src/Export/ProductImageService.php
```diff
@@ -32,16 +32,13 @@ public function __construct(RouterInterface $router)
...
+        if (!$productHasImages) {
             return $this->getFallbackImages();
         }
 
         $images = new ProductMediaCollection();
         if ($productHasImages) {
             $images = $this->getSortedProductImages($product);
-        } elseif ($hasVariantWithImages) {
-            $images = $this->getSortedVariantImages($product);
         }
```
TobiasGraml11: I guess this could make issues with the existing export logic. Does any test fail?
Maybe we need some `$considerVariants = true` parameter, which we set to `false` in our adapters.

src/Export/XmlExport.php
```diff
@@ -134,7 +138,37 @@ private function exportSingleItem(
...
+        /** @var ProductServiceSeparateVariants $productService */
+        $productService = $this->container->get(ProductServiceSeparateVariants::CONTAINER_ID);
+        $maxPropertiesCount = $productService->getMaxPropertiesCount($productEntity);
+        $pageSize = $this->calculatePageSize($maxPropertiesCount);
+        $initialItem = $this->xmlFileConverter->createItem($productEntity->getId());
+        $item = $exportItemAdapter->adapt($initialItem, $productEntity);
+
+        if ($item === null) {
+            $item = $initialItem;
+        }
```
TobiasGraml11: Why is this needed?
Is this, to export variant information, even when the main product throws an exception?

src/Export/XmlExport.php
```diff
@@ -134,7 +138,37 @@ private function exportSingleItem(
 
         /** @var ExportItemAdapter $exportItemAdapter */
         $exportItemAdapter = $this->container->get(ExportItemAdapter::class);
-        return $exportItemAdapter->adapt($this->xmlFileConverter->createItem($productEntity->getId()), $productEntity);
+        /** @var ProductServiceSeparateVariants $productService */
+        $productService = $this->container->get(ProductServiceSeparateVariants::CONTAINER_ID);
+        $maxPropertiesCount = $productService->getMaxPropertiesCount($productEntity);
+        $pageSize = $this->calculatePageSize($maxPropertiesCount);
+        $initialItem = $this->xmlFileConverter->createItem($productEntity->getId());
+        $item = $exportItemAdapter->adapt($initialItem, $productEntity);
```
TobiasGraml11: This whole block seems very intuitive, I have some ideas:
- Create member variable for `$exportItemAdapter` and `$productService` in this class and `$exportItemAdapter` in `ExportController`
- Forward both in `Export::getInstance`
- Move the `getMaxPropertiesCount` call inside of `calculatePageSize`
- Get the page size just one line above `buildVariantIterator`

src/Utils/Utils.php
```diff
@@ -94,24 +94,24 @@ public static function addProductAssociations(Criteria $criteria): Criteria
...
+                //                'children.manufacturer.translations',
+                //                'children.cover',
+                //                'children.properties',
+                //                'children.properties.group',
+                //                'children.categories',
+                //                'children.categories.seoUrls',
+                //                'children.properties.productConfiguratorSettings',
+                //                'children.properties.productConfiguratorSettings.option',
+                //                'children.properties.productConfiguratorSettings.option.group',
+                //                'children.properties.productConfiguratorSettings.option.group.translations',
```
TobiasGraml11: We will still need those associations for the old logic.
I would suggest to create `addChildrenAssociation`, and additionally call this in `ProductService::addProductAssociations`

tests/Export/Adapters/AttributeAdapterTest.php
```diff
@@ -5,6 +5,8 @@
 namespace FINDOLOGIC\FinSearch\Tests\Export\Adapters;
 
 use FINDOLOGIC\Export\Data\Attribute;
+use FINDOLOGIC\Export\Data\Ordernumber;
```
TobiasGraml11: Unused

tests/Export/Adapters/AttributeAdapterTest.php
```diff
@@ -68,13 +73,33 @@ protected function setUp(): void
...
+        ]);
+
+        $variantProduct = $this->createTestProduct([
+            'id' => $variantId,
+            'parentId' => $id,
+            'productNumber' => $variantProductNumber,
+            'shippingFree' => false
+        ]);
+
         $adapter = $this->getContainer()->get(AttributeAdapter::class);
```
TobiasGraml11: Should we set this as member variable in `setUp`?

TobiasGraml11: We can use that, when an adapter without custom config can be used

tests/Export/Adapters/AttributeAdapterTest.php
```diff
@@ -500,37 +546,40 @@ public function testOnlyUniqueCategoriesAreExported(): void
...
-            $this->salesChannelContext->getContext()
-        )->get($id);
-
         $config = $this->getMockedConfig();
-        $adapter = $this->getAttributeAdapter($config);
-        $attributes = $adapter->adapt($productEntity);
+        $initialItem = new XMLItem('123');
+        $exportItemAdapter = $this->getExportItemAdapter($config);
+
+        $item = $exportItemAdapter->adapt($initialItem, $productEntity);
```
TobiasGraml11, line 559: Can't we just use the `AttributeAdapter` here?

TobiasGraml11: This would also simplify the assertion without `ReflectionClass`

tests/Export/Adapters/OrderNumbersAdapterTest.php
```diff
@@ -30,16 +31,35 @@ protected function setUp(): void
...
-        $product = $this->createTestProduct();
 
-        $orderNumbers = $adapter->adapt($product);
+        $product = $this->createTestProduct([
+            'id' => $id
+        ]);
+
+        $variantProduct = $this->createTestProduct([
+           'id' => $variantId,
+            'parentId' => $id,
```
TobiasGraml11: You could remove this variable and use `$product->getId()`

tests/Export/Adapters/OrderNumbersAdapterTest.php
```diff
@@ -30,16 +31,35 @@ protected function setUp(): void
...
-        $adapter = $this->getContainer()->get(OrderNumberAdapter::class);
-        $product = $this->createTestProduct();
 
-        $orderNumbers = $adapter->adapt($product);
+        $product = $this->createTestProduct([
+            'id' => $id
+        ]);
+
+        $variantProduct = $this->createTestProduct([
+           'id' => $variantId,
```
TobiasGraml11: `$variantId` is not used otherwise, so you can just use `Uuid::randomHex()` directly.

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,559 @@
...
+use Shopware\Core\Content\Product\ProductCollection;
+use Shopware\Core\Content\Product\ProductEntity;
+use Shopware\Core\Content\Product\SalesChannel\ProductAvailableFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Dbal\Common\RepositoryIterator;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityRepository;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Aggregation\Bucket\TermsAggregation;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\AggregationResult\Bucket\TermsResult;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\EntitySearchResult;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\AndFilter;
```
TobiasGraml11: Unused

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,559 @@
...
+    }
+
+    public function searchVisibleProducts(
+        ?int $limit = null,
+        ?int $offset = null,
+        ?string $productId = null
+    ): EntitySearchResult {
+        $result = $this->getVisibleProducts($limit, $offset, $productId);
+
+        return $result;
```
TobiasGraml11:
```suggestion
        return $this->getVisibleProducts($limit, $offset, $productId);
```

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,559 @@
...
+
+        if ($productId) {
+            $this->addProductIdFilters($criteria, $productId);
+        }
+
+        return $this->container->get('product.repository')->search(
+            $criteria,
+            $this->salesChannelContext->getContext()
+        );
+    }
```
TobiasGraml11: I can't find any usage of this function

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,559 @@
...
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\NotFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\RangeFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Grouping\FieldGrouping;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\IdSearchResult;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Sorting\FieldSorting;
+use Shopware\Core\Framework\Uuid\Uuid;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+use Shopware\Core\System\SystemConfig\SystemConfigService;
+
+class ProductServiceSeparateVariants
```
TobiasGraml11, line 34: I always had a hard time finding the correct place of functions, while working on it.
Could you sort them a bit by the call timing? As an example, `getVisibleProducts` is basically one of the entry point functions, but is placed very far down.
Also: `public` functions should usually be placed on top and all the others afterwards

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,559 @@
...
+        return $this->container->get('product.repository')->search(
+            $criteria,
+            $this->salesChannelContext->getContext()
+        );
+    }
+
+    public function buildVariantIterator(ProductEntity $product, int $pageSize): RepositoryIterator
+    {
+        $criteria = new Criteria();
+        $productRepository = $this->container->get('product.repository');
```
TobiasGraml11: The product repository is used 7 times in the class, we should fetch this once in the constructor.

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,559 @@
...
+        }
+
+        if (empty($realProductIds)) {
+            return null;
+        }
+
+        return $this->getRealMainVariants($realProductIds);
+    }
+
+    protected function getFirstChildren(ProductEntity $product): ?ProductEntity
```
TobiasGraml11: `getFirstChild` seems to be correct, according to the functionality

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,559 @@
...
+            return $this->getCheapestProducts($products->getEntities());
+        }
+
+        $mainVariants = $this->getConfiguredMainVariants($products->getEntities());
+
+        if (!$mainVariants) {
+            return $products;
+        }
+
+        return $mainVariants;
```
TobiasGraml11:
```suggestion
        $mainVariants = $this->getConfiguredMainVariants($products->getEntities());
        
        return $mainVariants ?: $products;
```

TobiasGraml11: Wouldn't this wrongly export the main variant for "Shopware default", or is this how it should be anyway?

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,559 @@
...
+                continue;
+            }
+
+            $cheapestVariantPrice = $cheapestVariant->getCurrencyPrice($currencyId);
+
+            if ($productPrice->getGross() === 0.0) {
+                $realCheapestProduct = $cheapestVariant;
+            } else {
+                $realCheapestProduct = $productPrice->getGross() <= $cheapestVariantPrice->getGross()
+                    ? $product : $cheapestVariant;
```
TobiasGraml11:
```suggestion
                $realCheapestProduct = $productPrice->getGross() <= $cheapestVariantPrice->getGross()
                    ? $product
                    : $cheapestVariant;
```

src/Export/ProductServiceSeparateVariants.php
```diff
@@ -0,0 +1,559 @@
...
+            [
+                new EqualsFilter('parentId', null),
+                new EqualsFilter('active', true)
+            ]
+        );
+
+        /**
+         * We still need to fetch the product if it is not active, but have children products.
+         */
+        $inactiveParentWithChildsFilter = new MultiFilter(
```
TobiasGraml11:
```suggestion
        $inactiveParentWithChildrenFilter = new MultiFilter(
```

TobiasGraml11 (changes requested):

src/Export/XmlExport.php
msilinskas: Yes, it's for exporting a variants information, even if the main variant throws an error

src/Export/ProductServiceSeparateVariants.php
msilinskas: It's used in the ProductServiceSeparateVariantsTest::testFindsAllProducts

tests/Export/Adapters/AttributeAdapterTest.php
msilinskas: Calling the attribute adapter here directly we can't make sure, that only unique categories was exported

src/Export/ProductServiceSeparateVariants.php
msilinskas: In the old behavior we also used a main variants if it was able to.

src/Export/ProductServiceSeparateVariants.php
TobiasGraml11: Yeah, but why would we need to have a function, just to test it?
So if the function is not needed, the test is also irrelevant.

tests/Export/Adapters/AttributeAdapterTest.php
TobiasGraml11: Ah, I see!

src/Export/Adapters/AdapterFactory.php
```diff
@@ -83,6 +87,7 @@ public function __construct(
         $this->summaryAdapter = $summaryAdapter;
         $this->urlAdapter = $urlAdapter;
         $this->userGroupsAdapter = $userGroupsAdapter;
+        $this->shopwarePropertiesAdapter = $shopwarePropertiesAdapter;
```
TobiasGraml11: Please keep the alphabetic order

src/Export/Adapters/AdapterFactory.php
```diff
@@ -159,4 +164,9 @@ public function getUserGroupsAdapter(): UserGroupsAdapter
     {
         return $this->userGroupsAdapter;
     }
+
+    public function getShopwarePropertiesAdapter(): ShopwarePropertiesAdapter
+    {
+        return $this->shopwarePropertiesAdapter;
+    }
```
TobiasGraml11: Please keep the alphabetic order

src/Export/Adapters/PropertiesAdapter.php
```diff
@@ -14,7 +14,7 @@
 use Shopware\Core\System\SalesChannel\SalesChannelContext;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
-class PropertiesAdapter
+class PropertiesAdapter extends ShopwarePropertiesAdapter
```
TobiasGraml11: Please also split the adapter tests and move the tests to the correct test file

TobiasGraml11: Please make them independent from each other, so call each one separately.

TobiasGraml11: To make it more clear, I would call `PropertiesAdapter` something like `DefaultPropertiesAdapter`

src/Resources/config/services/services.xml
```diff
@@ -86,6 +86,7 @@
             <argument type="service" id="FINDOLOGIC\FinSearch\Export\Adapters\SummaryAdapter" />
             <argument type="service" id="FINDOLOGIC\FinSearch\Export\Adapters\UrlAdapter" />
             <argument type="service" id="FINDOLOGIC\FinSearch\Export\Adapters\UserGroupsAdapter" />
+            <argument type="service" id="FINDOLOGIC\FinSearch\Export\Adapters\ShopwarePropertiesAdapter" />
```
TobiasGraml11: Please keep the alphabetic order

src/Resources/config/services/services.xml
```diff
@@ -124,6 +125,12 @@
             <argument type="service" id="FINDOLOGIC\FinSearch\Struct\Config" />
         </service>
 
+        <service id="FINDOLOGIC\FinSearch\Export\Adapters\ShopwarePropertiesAdapter" public="true">
+            <argument type="service" id="fin_search.sales_channel_context" />
+            <argument type="service" id="Shopware\Core\Framework\Adapter\Translation\Translator" />
+            <argument type="service" id="FINDOLOGIC\FinSearch\Struct\Config" />
+        </service>
+
```
TobiasGraml11: Please keep the alphabetic order

TobiasGraml11 (changes requested):

TobiasGraml11 (approved):

PR #284, opened 2022-05-17
SW-700 Do not hide vendor filter for vendor search pages

TobiasGraml11:


src/Resources/views/storefront/component/listing/filter/filter-property-select.html.twig
```diff
@@ -25,7 +25,16 @@
               {% endif %}
         >
             {% block component_filter_property_select_checkox_input %}
-                {{ parent() }}
+                <input type="checkbox"
+                       class="filter-property-select-checkbox filter-multi-select-checkbox"
+                       id="{{ element.uuid }}"
```
TobiasGraml11, line 30: Changed from the SW implementation `id="{{ id }}"`

dblaichinger (approved):

PR #291, opened 2022-06-15
SW-730 Split product service into multiple services

TobiasGraml11:


src/Controller/ExportController.php
```diff
@@ -250,11 +251,18 @@ private function buildExportContext(): ExportContext
...
             Utils::fetchNavigationCategoryFromSalesChannel(
                 $this->container->get('category.repository'),
                 $this->salesChannelContext->getSalesChannel()
             )
         );
     }
+
+    private function getAllCustomerGroups(): array
+    {
+        return $this->container->get('customer_group.repository')
```
msilinskas: wouldn't it better to add this to the constructor instead of getting it from the container?

PR #299, opened 2022-06-29
SW-508 it must be possible to set configurations for all sales channels

TobiasGraml11:


TobiasGraml11:
"Code linting (PHP 7.3)" check also stuck for new PR

PR #304, opened 2022-07-15
element.uuid is null

hbaeumer:
<input id ="" …>
is not valid.
label reference is not valid.


TobiasGraml11:
Hello @hbaeumer!

Thanks for your PR, could you please give me an example, where this UUID is not set?
This one is required, so we can support the same filter values inside different filters.

image

Cheers,
Tobias

TobiasGraml11:
Closing this due to inactivity

PR #307, opened 2022-08-22
SW-740 Filter valid SEO Urls

nenadj-soprex:


src/Export/UrlBuilderService.php
```diff
@@ -123,7 +123,7 @@ protected function getSalesChannelDomain(): ?string
      */
     protected function getProductSeoPath(ProductEntity $product): ?string
     {
-        $allSeoUrls = $product->getSeoUrls();
+        $allSeoUrls = $this->removeInvalidUrls($product->getSeoUrls()->getElements());
         if (!$allSeoUrls) {
```
TobiasGraml11: This `null` check is still needed and has no purpose in the current state

src/Export/UrlBuilderService.php
```diff
@@ -142,6 +142,20 @@ protected function getProductSeoPath(ProductEntity $product): ?string
...
+    public function removeInvalidUrls(array $seoUrls): SeoUrlCollection
+    {
+        foreach ($seoUrls as $key => $seoUrl) {
+            $urlStr = $seoUrl->getSeoPathInfo();
+            if (!filter_var(sprintf('https://dummy.com%s"', $urlStr), FILTER_VALIDATE_URL)) {
+                unset($seoUrls[$key]);
+            }
+        }
+        return new SeoUrlCollection($seoUrls);
+    }
```
TobiasGraml11:
```suggestion
        public function removeInvalidUrls(SeoUrlCollection $seoUrls): SeoUrlCollection
    {
        return $seoUrls->filter(function (SeoUrlEntity $seoUrl) {
            return filter_var(
                sprintf('https://dummy.com%s"', $seoUrl->getSeoPathInfo()),
                FILTER_VALIDATE_URL
            );
        });
    }
```

tests/Export/UrlBuilderServiceTest.php
```diff
@@ -0,0 +1,64 @@
...
+        foreach ($seoPathInfos as $seoPathInfo) {
+            $seoUrlEntity = new SeoUrlEntity();
+            $seoUrlEntity->setSeoPathInfo($seoPathInfo);
+            $seoUrlEntity->setId($seoPathInfo);
+            array_push($seoUrlCollection, $seoUrlEntity);
+        }
+        $allowedUrls = $this->urlBuilderService->removeInvalidUrls($seoUrlCollection);
+
+        $this->assertSame(2, $allowedUrls->count());
+    }
```
TobiasGraml11:
```suggestion
    public function testRemoveInvalidUrls(): void
    {
        $expectedUrlCount = 2;
        $seoPathInfos = [
            '/failed seo url with spaces',
            'failedSeoUrlWithoutSlash',
            '/correctSeoUrl-One',
            '/correctSeoUrlTwo'
        ];

        $seoUrlCollection = new SeoUrlCollection();
        foreach ($seoPathInfos as $seoPathInfo) {
            $seoUrlEntity = new SeoUrlEntity();
            $seoUrlEntity->setId(Uuid::randomHex());
            $seoUrlEntity->setSeoPathInfo($seoPathInfo);

            $seoUrlCollection->add($seoUrlEntity);
        }

        $this->assertSame(
            $expectedUrlCount,
            $this->urlBuilderService->removeInvalidUrls($seoUrlCollection)->count()
        );
    }
```

src/Export/UrlBuilderService.php
TobiasGraml11: So you need to:
- Still check if `$product->getSeoUrls()` returns something
- Additionally check if `removeInvalidUrls` does return an empty collection

TobiasGraml11 (changes requested):

src/Export/UrlBuilderService.php
nenadj-soprex: @TobiasGraml11 
Guess only one condition is needed?
Then it is enough?: `if (!count($allSeoUrls))`
Not needed to check productSeoUrls too:`if (!count($product->getSeoUrls()) || !count($allSeoUrls))`

TobiasGraml11: You will need something like this:
```php
if (!$product->getSeoUrls()) {
    return null;
}

$allSeoUrls = $this->removeInvalidUrls($product->getSeoUrls())

if (!$allSeoUrls->count()) {
    return null;
}
```

nenadj-soprex: We also need to check count() on $product->getSeoUrls(), correct?

```
if (!$product->getSeoUrls()->count()) {
    return null;
}
```

TobiasGraml11: No, because the code will not throw an exception in that case, this will be handled by the check for `$allSeoUrls`

TobiasGraml11 (approved):

PR #308, opened 2022-08-23
SW-746 Use first element of seoUrls if canonicalSeoUrls is null

nenadj-soprex:


tests/Utils/UtilsTest.php
```diff
@@ -399,4 +399,12 @@ public function testCategoryPathIsProperlyBuiltWhenMainCategoryIsInADeeperPath()
...
         $this->assertSame($expectedCategoryPath, $categoryPath);
     }
+
+    public static function callMethod($obj, $name, array $args)
+    {
+        $class = new \ReflectionClass($obj);
+        $method = $class->getMethod($name);
+        $method->setAccessible(true);
+        return $method->invokeArgs($obj, $args);
+    }
```
TobiasGraml11: This file is not a utils class for the tests, it's tests for the existing `Utils` class. This will need to be placed in a separate helper file.

tests/Export/UrlBuilderServiceTest.php
```diff
@@ -44,28 +45,132 @@ protected function setUp(): void
...
-            $this->urlBuilderService->removeInvalidUrls($seoUrlCollection)->count()
+            UtilsTest::callMethod(
+                $this->urlBuilderService,
+                'removeInvalidUrls',
+                [$seoUrlCollection]
+            )->count()
+        );
+    }
+
+    public function removeInvalidUrlsProvider(): array
```
TobiasGraml11, line 48: Please always define the provider above the test, that uses it.

tests/Export/UrlBuilderServiceTest.php
```diff
@@ -44,28 +45,132 @@ protected function setUp(): void
...
+        $seoUrl = UtilsTest::callMethod(
+            $this->urlBuilderService,
+            'getProductSeoPath',
+            [$product]
         );
+
+        $this->assertSame($expectedSeoUrl, $seoUrl);
+    }
+
+    public function productSeoPathProvider(): array
```
TobiasGraml11, line 103: Please always define the provider above the test, that uses it.

TobiasGraml11 (changes requested):

TobiasGraml11 (approved):

PR #309, opened 2022-08-24
SW-737 Filter values of color picker should be selectable by clicking on the value next to the tile

bojand-soprex:


src/Resources/views/storefront/component/listing/filter/filter-property-select.html.twig
```diff
@@ -27,7 +27,7 @@
             {% block component_filter_property_select_checkox_input %}
                 <input type="checkbox"
                        class="filter-property-select-checkbox filter-multi-select-checkbox"
-                       id="{{ element.uuid }}"
+                       id="{{ element.id }}"
```
TobiasGraml11: I remember that this `uuid` was needed for filters with the same filter values: https://findologic.myjetbrains.com/youtrack/issue/SW-451. Because of that, I do not want to change it, as it could re-introduce issues with some edge cases.


https://github.com/shopware/platform/blob/trunk/src/Storefront/Resources/views/storefront/component/listing/filter/filter-property-select.html.twig
Please see my suggestion below, on how we could get the correct `for` property instead:

image

TobiasGraml11 (changes requested):

PR #312, opened 2022-09-05
SW-749 Upgrade code level to PHP 7.4

TobiasGraml11:


src/Core/Content/Category/SalesChannel/CachedCategoryRoute.php
```diff
@@ -44,13 +38,13 @@ public function getDecorated(): AbstractCategoryRoute
...
         Request $request,
-        SalesChannelContext $salesChannelContext
+        SalesChannelContext $context
     ): CategoryRouteResponse {
-        $this->initializePluginState($salesChannelContext, $request);
+        $this->initializePluginState($context, $request);
 
-        return $this->getDecorated()->load($categoryId, $request, $salesChannelContext);
+        return $this->getDecorated()->load($navigationId, $request, $context);
     }
```
TobiasGraml11, line 48: Argument names were updated in `AbstractCategoryRoute`

src/Findologic/Response/ResponseParser.php
```diff
@@ -21,20 +21,11 @@
...
-     * @var ServiceConfigResource|null
-     */
-    protected $serviceConfigResource;
+    protected ?ServiceConfigResource $serviceConfigResource;
 
-    /**
-     * @var Config
-     */
-    protected $config;
+    protected ?Config $config;
```
TobiasGraml11, line 28: This was wrongly typed, as the constructor argument is also nullable.

src/Findologic/Response/Xml21/Filter/ColorPickerFilter.php
```diff
@@ -6,6 +6,5 @@
 
 class ColorPickerFilter extends Filter
 {
-    /** @var string */
-    protected $displayType = 'color';
+    protected ?string $displayType = 'color';
```
TobiasGraml11, line 9: Type needs to be the same as the parent, therefore I set it nullable

src/Findologic/Config/FinSearchConfigEntity.php
```diff
@@ -13,35 +13,18 @@ class FinSearchConfigEntity extends Entity
...
-     */
-    protected $configurationKey;
-
-    /**
-     * @var mixed
-     */
+    protected string $configurationKey;
+
+    /** @var mixed */
     protected $configurationValue;
```
msilinskas, line 19: Couldn't you declare types for this variable?

TobiasGraml11: It can basically be any type, that's why I kept using `mixed`

TobiasGraml11: `mixed` is not available as variable type and union types are still not supported (`array|int|bool|string`)

PR #323, opened 2022-10-14
SW-763 Export cat_urls for API integrations

TobiasGraml11:


src/Struct/FindologicProduct.php
```diff
@@ -856,7 +856,7 @@ protected function setCategoriesAndCatUrls(): void
             $this->parseCategoryAttributes($dynamicGroupCategories, $catUrls, $categories);
         }
 
-        if ($this->isDirectIntegration() && !Utils::isEmpty($catUrls)) {
```
nenadj-soprex, line 859: We have one more condition for "cat_urls" in this file at line 1098

image

nenadj-soprex: @TobiasGraml11 if we remove this call of `isDirectIntegration()` at line 1098 line, we can remove and definition of function itself, because it is no longer used, lines 1107 - 1113.

TobiasGraml11: We need to keep `isDirectIntegration()`, in case existing (legacy) extensions still use the function.

PR #333, opened 2022-11-17
SW-772 Export variant options

TobiasGraml11:


TobiasGraml11:
Fixed with version bump in #334

PR #342, opened 2022-12-19
SW-780 Fix compatibility check

TobiasGraml11:


.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 env:
   PLUGIN_NAME: FinSearch
-  PLUGIN_MIN_REQUIREMENT: ^6.4.6
+  PLUGIN_MIN_REQUIREMENT: ~6.4.6
```
TobiasGraml11, line 11: We do not want to support `6.5` without manually defining it, as there will probably be breaking changes.

bojand-soprex (approved):
The code LGTM as well the logic for the new changes.

PR #344, opened 2022-12-21
SW-777 Round to two decimals minTotal, minSelected and filterMin on RangeSlider

nenadj-soprex:


src/Findologic/Response/Xml21/Filter/Filter.php
```diff
@@ -108,14 +108,14 @@ private static function handleRangeSliderFilter(ApiRangeSliderFilter $filter): R
 
         if ($filter->getAttributes()->getTotalRange()) {
             $customFilter->setTotalRange([
-                self::FILTER_RANGE_MIN => $filter->getAttributes()->getTotalRange()->getMin(),
+                self::FILTER_RANGE_MIN => round($filter->getAttributes()->getTotalRange()->getMin(), 2),
```
TobiasGraml11: Shouldn't we also round the max value?

nenadj-soprex: From the my first investigation, we have a problem exclusively with zero decimals... Give me one more try to test again big values. If it is true, and we have only issue with zero values, under one, maybe we can only need to hande min values. Or how you say :)

TobiasGraml11: I think it would generally make sense, just to be consistent here with the values.

src/Findologic/Response/Xml21/Filter/Filter.php
```diff
@@ -108,15 +108,15 @@ private static function handleRangeSliderFilter(ApiRangeSliderFilter $filter): R
 
         if ($filter->getAttributes()->getTotalRange()) {
             $customFilter->setTotalRange([
-                self::FILTER_RANGE_MIN => $filter->getAttributes()->getTotalRange()->getMin(),
-                self::FILTER_RANGE_MAX => $filter->getAttributes()->getTotalRange()->getMax(),
+                self::FILTER_RANGE_MIN => round($filter->getAttributes()->getTotalRange()->getMin(), 2),
+                self::FILTER_RANGE_MAX => round($filter->getAttributes()->getTotalRange()->getMax(), 2),
```
TobiasGraml11: I think it could be an issue, when using `round()` for both. We probably need to use `floor()` for the min values and `ceil()` for the max values.
If we round the min value 0.655 to 0.66, we automatically kick out all the products with 0.655 from the range. Rounding to 0.65 would keep the value in the range.

TobiasGraml11 (changes requested):

src/Findologic/Response/Xml21/Filter/RangeSliderFilter.php
```diff
@@ -16,7 +16,7 @@ class RangeSliderFilter extends Filter
 
     private ?float $step = null;
 
-    private string $unit = '€';
+    private string $unit = '';
```
TobiasGraml11, line 19: If no unit is configured, there should also be no unit shown

src/Resources/views/storefront/component/listing/filter/filter-slider-range.html.twig
```diff
@@ -1,7 +1,7 @@
 {% set filterItemId = 'filter-' ~ name ~ '-' ~ random() %}
 
 {% if currencySymbol is not defined %}
-    {% set currencySymbol = page.header.activeCurrency.symbol %}
+    {% set currencySymbol = '' %}
```
TobiasGraml11: If no unit is configured, there should also be no unit shown

src/Findologic/Response/Xml21/Filter/Filter.php
nenadj-soprex: Fixed by your solution via javascript

TobiasGraml11 (approved):

PR #349, opened 2023-01-18
SW-788 The filter request handling by configuration on Category Pages

nenadj-soprex:


src/Storefront/Controller/CmsController.php
```diff
@@ -116,19 +116,20 @@ public function category(
...
+        $event = new ProductListingCriteriaEvent($request, new Criteria(), $salesChannelContext);
         $this->config->initializeBySalesChannel($salesChannelContext);
         if (
             !Utils::shouldHandleRequest(
                 $request,
                 $salesChannelContext->getContext(),
                 $this->serviceConfigResource,
-                $this->config
+                $this->config,
+                !($event instanceof ProductSearchCriteriaEvent)
```
TobiasGraml11: 1. This route is within `/navigation/{navigationId}`, so I don't expect this to ever be about search
2. You manually create the event as `ProductListingCriteriaEvent`, so this will always be `true`

Just hardcode this with `true` and move the `$event` initialization back to after the if block.

TobiasGraml11 (changes requested):

TobiasGraml11 (approved):

PR #352, opened 2023-03-16
SW-791 Update extended templates

TobiasGraml11:


nenadj-soprex (commented):
@TobiasGraml11 Code looks good to me.

PR #363, opened 2023-04-13
SW-797 Use variants logic within search results

TobiasGraml11:


nenadj-soprex (approved):
@TobiasGraml11 Code looks good to me.

PR #376, opened 2023-08-07
SW-802 Make sure the plugin works for 6.5.3.x and 6.5.4.x

TobiasGraml11:


.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 env:
   PLUGIN_NAME: FinSearch
-  PLUGIN_MIN_REQUIREMENT: ^6.5
+  PLUGIN_MIN_REQUIREMENT: ^6.5.3
```
TobiasGraml11, line 11: We decided to drop the support for lower versions, because of the change of Shopware

PR #387, opened 2024-03-11
SW-812_search_doesnt_work_when_findologic_isnt_active_for_sales_channel

Radomir98:


src/Core/Content/Product/SalesChannel/Search/ProductSearchRoute.php
```diff
@@ -85,12 +85,15 @@ public function load(
...
         if (!$shouldHandleRequest) {
-            return $this->decorated->load($request, $context, $criteria);
+            $result = $this->decorated->load($request, $context, $criteria)->getListingResult();
         }
 
         $query = $request->query->get('search');
-        $result = $this->doSearch($criteria, $context, $query);
-        $result = ProductListingResult::createFrom($result);
+
+        if (!isset($result)) {
```
TobiasGraml11: Wouldn't it make more sense to add it within an else block above?

TobiasGraml11 (changes requested):

TobiasGraml11 (approved):

PR #388, opened 2024-03-13
Bump node version to 18.18.0

bojand-soprex:


TobiasGraml11 (changes requested):
The linting is still failing with the same issue

TobiasGraml11 (approved):

PR #399, opened 2024-04-24
SW-805 Wrong prices are exported

Radomir98:


src/Resources/app/administration/src/module/findologic-module/snippet/en-GB.json
```diff
@@ -91,6 +91,10 @@
                         "label": "Price when buying one unit"
                     }
                 },
+                "domain": {
+                    "label": "Domain for generating product url",
+                    "helpText": "This domain is used for multi-domain sales channels, and the selected domain will be used as uri for product url"
```
TobiasGraml11: Please adapt these, so they represent the real usage. For now, we basically only use the configured currency. I'll translate them to German afterwards.

src/Findologic/Config/FindologicConfigService.php
```diff
@@ -277,4 +280,25 @@ private function getConfigValues(array $configValues, array $keys, $value): arra
 
         return $configValues;
     }
+
+    public function getDomainId($channelId = null, $languageId = null): ?string
```
TobiasGraml11: As you can see by the existing function, this service is not really the place to fetch specific config values. They are only fetched by the config key.

src/Export/Services/SalesChannelService.php
```diff
@@ -39,6 +41,7 @@ public function getSalesChannelContext(
             (new Criteria())->addFilter(new EqualsFilter('configurationKey', 'FinSearch.config.shopkey')),
             $currentContext->getContext()
         );
+        $currencyId = $this->findologicConfigService->getDomainCurrencyId($currentContext);
```
TobiasGraml11: Relying on the current context could lead to issues, when the shopkey is requested on a different sales channel.
We should fetch the domain within the if block below according to the `languageId` and `salesChannelId`. If no `domainId` is configured, I guess we just take the first one.

TobiasGraml11 (changes requested):

Bibarella (approved):
LGTM

PR #406, opened 2024-08-16
SW-817 Pagination in categories have to work

Radomir98:


Bibarella (changes requested):
There are nested if statements that could be flattened for better readability. Might be good to use early return or break the logic into separate methods.  Utils::shouldHandleRequest method is called twice with almost identical parameters, except for the last boolean parameter in the elseif block. This introduces unnecessary repetition.

Consider extracting the common logic or calling Utils::shouldHandleRequest once, where the additional boolean can be handled separately if required. Store the result of Utils::shouldHandleRequest in a variable if the condition is the same, and reuse it where applicable.

Consider extracting the search and navigation logic into their own methods. Would make the main method cleaner and adheres to the SRP. 

Suggestion:
`$this->config->initializeBySalesChannel($context);

if (Utils::isSearchPage($request)) {
    if ($this->shouldHandleRequest($request, $context)) {
        $this->findologicSearchService->doSearch($request, $criteria, $context);
    }
} elseif (Utils::isNavigationPage($request)) {
    if ($this->shouldHandleRequest($request, $context, true)) {
        $this->findologicSearchService->doNavigation($request, $criteria, $context);
    }
}

private function shouldHandleRequest($request, $context, $isNavigation = false) {
    return Utils::shouldHandleRequest(
        $request,
        $context->getContext(),
        $this->serviceConfigResource,
        $this->config,
        $isNavigation
    );
}`

Bibarella (approved):
LGTM

